PR #13, opened 2015-01-05
include VCF Header info in output from VCFRecordReader

eastlondoner:
This fixes an issue where it is impossible to access information contained in the VCF header within Map/Reduce methods because the VCFRecordReader does not set the header information on the VCFContextWritables that it produces.
The VCFRecordReader class does read the header (and throws an error if it is not present) because it needs it to deal with contigs. Passing the information out is a very simple change.


AndreSchumacher:
Thanks a lot @eastlondoner and apologies it took so long.

eastlondoner:
No worries!

PR #17, opened 2015-05-11
Generic output stream for {Fastq,Qseq}OutputFormat

ilveroluca:
This pull request changes the Fastq and Qseq output format classes to require an `OutputStream` object rather than a `DataOutputStream` (which wasn't really needed).  The more general output type makes the code more re-usable and easier to test.


AndreSchumacher:
Looks good to me, mind merging that in yourself @ilveroluca ?

ilveroluca:
done and done!

PR #20, opened 2015-07-14
Fallback to probabilistic splits if index is bad.

trozamon:
BAMInputFormat throws an exception if the generated index is somehow
incorrect. Now, BAMInputFormat falls back to generating probabilistic
splits instead of throwing an exception.

I am using Hadoop-BAM to run summary statistics the entire 1000 genomes dataset as a proof of concept. I accidentally generated a handful of the indices incorrectly, and it is much more convenient to have BAMInputFormat fall back to probabilistic input splitting than hard fail when 10 out of 2500 ".splitting-bai" files are incorrect.


AndreSchumacher:
Makes sense. @trozamon thanks for the PR!

PR #22, opened 2015-08-24
set SortOrder before header on SAMTextWriter

### LINKED ISSUE #21: SAMTextWriter.setSortOrder is not called before SAMTextWriter.setHeader in SAMRecordWriter
[`SAMTextWriter.setSortOrder` is not called before `SAMTextWriter.setHeader` in `SAMRecordWriter.init`](https://github.com/ryan-williams/Hadoop-BAM/commit/97e523ecb734badc8fe0ffe984d6efe0b9b60de3), despite [this comment stating that it must](https://github.com/samtools/htsjdk/blob/1.118/src/java/htsjdk/samtools/SAMFileWriterImpl.java#L82) (link is from htsjdk 1.118, but this API and comment [still exist in htsjdk 1.138](https://github.com/samtools/htsjdk/blob/1.138/src/java/htsjdk/samtools/SAMFileWriterImpl.java#L83)).

I noticed this because all `.sam` output files in ADAM have `@HD` lines with a `SO:unsorted` tag, [even if we set a `coordinate` sort on the `SAMFileHeader` before writing](https://github.com/bigdatagenomics/adam/blob/de3fa55b95069fcf8d8f4096f5bf7f0a3fb785dd/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala#L248).

ryan-williams:
fixes #21

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 .idea
 *.iml
 maven-metadata-local.xml
+*~
--- a/src/main/java/org/seqdoop/hadoop_bam/SAMRecordWriter.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/SAMRecordWriter.java
@@ -90,6 +90,7 @@ public abstract class SAMRecordWriter<K>
 		this.header = header;
 		writer = new SAMTextWriter(output);
 
+		writer.setSortOrder(header.getSortOrder(), false);
 		if (writeHeader)
 			writer.setHeader(header);
 	}


ilveroluca:
Thanks for the patch!

laserson:
Just ran into this.  Any chance you could cut a point release?

AndreSchumacher:
Yes, we'll do that, hopefully this weekend.

PR #25, opened 2015-10-21
Fix BAMSplitGuesser to check that read name is of length one or more.

tomwhite:
This fixes a bug when reading BAM files without an index that was found in https://github.com/broadinstitute/gatk/issues/1005. The problem is that the heuristics in BAMSplitGuesser are not strong enough and in some cases can return a position that is not the start of a BAM record. The test in this PR fails for the existing BAMSplitGuesser; it passes with the fix.


ilveroluca:
Thanks Tom.

Issue #24, opened 2015-10-20
Any plans for support for CRAM files ?


cmnbroad:
Or has anyone done any work on what it would take ?

tomwhite:
I've had a brief look at what it would take to read CRAM files and I've posted some initial code here: https://github.com/HadoopGenomics/Hadoop-BAM/compare/master...tomwhite:cram (still needs tests).

The basic idea is to scan through the CRAM file to find all the container offsets. This should be fairly efficient since it is essentially a seek and a short read (to read the container header) for each container. This can be done in the implementation of InputFormat#getSplits, but if this turns out to be slow then it would be possible to write an index file, just like BAMInputFormat does.

Once the container offsets have been found then the file splits can be adjusted to align with the container boundaries. (So one split corresponds to one or more CRAM containers.) Then the record reader only needs to read the CRAM file header, before it reads the records from the containers in its split.

When reading CRAM files you need to have a reference file available. Currently, I've assumed it's on the local file system, but we really need something like https://github.com/samtools/htsjdk/pull/308 so the reference can be read from HDFS.

cmnbroad:
Thx - I'll have a look.

tomwhite:
Initial support was merged in https://github.com/HadoopGenomics/Hadoop-BAM/issues/28, with https://github.com/HadoopGenomics/Hadoop-BAM/issues/31 as a follow up. This issue can be closed now.

PR #28, opened 2015-11-09
Add CRAMInputFormat.

tomwhite:
This is an implementation of CRAMInputFormat for reading CRAM files. Includes tests. See #24 for a brief description of how it works.

The main restriction is that the reference sequence (needed for reading CRAM) needs to be available on the local filesystem. This restriction will be removed once https://github.com/samtools/htsjdk/pull/308 is available in htsjdk.


AndreSchumacher:
@tomwhite first of all thanks a lot for looking into CRAM support. Just two comments:

a) the splitting by containers: as we briefly discussed in the last meeting, is it actually clear that this partitioning is fine enough for allow for a reasonable degree of parallelism? If so, great, but it would be good to have some idea say what's a typical ratio of container size to CRAM file size or something.

b) the limitation of the reference to be a java.io.File (or even java.nio.file.Path) is I think a big one. That should not be a blocker for this PR but I think it would be much better if htsjdk's ReferenceSource would take an InputStream or SeekableStream or something instead of path. Here we already rely on HDFS for various other things so I think it's reasonable to assume people just want to dump their fasta files to HDFS / S3 (and that allows for local paths via "file://path" as well). Maybe I am missing something but the library you linked to (https://github.com/damiencarol/jsr203-hadoop) is still in beta?

tomwhite:
Thanks for taking a look at this Andre.

>  is it actually clear that this partitioning is fine enough for allow for a reasonable degree of parallelism?

The CRAM spec recommends containers of 1MB (see section 13.4 of http://samtools.github.io/hts-specs/CRAMv3.pdf). This small enough compared to the total size of a CRAM file (gigabytes) to allow for a reasonable degree of parallelism.  Also, assuming HDFS blocks of 128MB (the default), container of 1MB means that <1% of each split will be a non-local read, so this would be fine from a locality point of view too. 

> the limitation of the reference to be a java.io.File (or even java.nio.file.Path) is I think a big one

We definitely need to remove that limitation, but it needs more work in htsjdk. The java.nio.file.Path route is probably what we need long term (the jsr203-hadoop library is beta, but works well enough for the basic operations we need). Using SeekableStream instead of Path is tricky since the reference path is used to look for the name of the reference index, however with a bit more work it should be possible to specify the index as an InputStream too.

Regardless, this PR should not wait for the htsjdk to be settled, since we can remove the limitation in the future.

PR #29, opened 2015-11-13
Improve BAMSplitGuesser to check that CIGAR codes are within range.

tomwhite:
This fixes the bug found by David Adams (https://github.com/broadinstitute/gatk/issues/1098). The BAM file that exposed this is many GB in size so it can't be checked in here. However, when I tested with this fix the error no longer occurred.


tomwhite:
This fix is simpler than the one I originally had, since it just calls `record.getCigar()` to force decoding once a candidate start record position is found. It would be straightforward to add other calls to force decoding of other fields (as long as they don't reference header values), but this change is sufficient to fix the current bug.

AndreSchumacher:
@tomwhite thanks! Great that the fix was so simple.

PR #38, opened 2015-11-30
Fix htsjdk deprecation warnings.

tomwhite:


src/main/java/org/seqdoop/hadoop_bam/BCFRecordWriter.java
```diff
@@ -106,8 +104,8 @@ private void init(
 		final BCFStoppableOutputStream stopOut =
 			new BCFStoppableOutputStream(!writeHeader, output);
 
-		writer = VariantContextWriterFactory.create(
-			stopOut, null, EnumSet.of(Options.FORCE_BCF));
+		writer = new VariantContextWriterBuilder().clearOptions()
+				.setOutputBCFStream(stopOut).build();
```
AndreSchumacher: It seems this does not actually set the FORCE_BCF option? As far as I see it probably will not matter much but in order to not introduce any side effects should we maybe set it via setOption?

AndreSchumacher:
@tomwhite thanks for cleaning up those warnings. I added one comment above. Otherwise looks good to me, feel free to merge.

src/main/java/org/seqdoop/hadoop_bam/BCFRecordWriter.java
tomwhite: Fixed.

tomwhite:
Thanks for taking a look @AndreSchumacher!

PR #40, opened 2015-12-04
bump htsjdk to 2.0.0 for better cram support. Pass validation stringency to CRAMRecordReader. Java8

akiezun:
in GATK (https://github.com/broadinstitute/gatk), we require htsjdk 2.0.0 for CRAM support. That is incompatible with hadoop-bam because API has changed (requires additional parameter in CRAMRecordReader). This change is to push the htsjdk version to 2.0.0 and fix the API issue. 

Java version needs to move to 8 otherwise blows up because htsjdk is on Java 8. Java 8 is a superior language compared to 1.6 and plays nicely with Spark and is mature/stable now so I think migration should not be a issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
 language: java
 jdk:
   - oraclejdk8
-  - oraclejdk7
-  - openjdk7
-  - openjdk6
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,7 @@
     <url>http://sourceforge.net/projects/hadoop-bam/</url>
 
     <properties>
-        <java.version>1.6</java.version>
+        <java.version>1.8</java.version>
         <htsjdk.version>2.0.0</htsjdk.version>
         <!-- CHANGE THIS FOR A DIFFERENT VERSION OF HADOOP -->
         <hadoop.version>2.2.0</hadoop.version>
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
 
     <properties>
         <java.version>1.6</java.version>
-        <htsjdk.version>1.140</htsjdk.version>
+        <htsjdk.version>2.0.0</htsjdk.version>
         <!-- CHANGE THIS FOR A DIFFERENT VERSION OF HADOOP -->
         <hadoop.version>2.2.0</hadoop.version>
         <!--
--- a/src/main/java/org/seqdoop/hadoop_bam/CRAMInputFormat.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/CRAMInputFormat.java
@@ -3,11 +3,6 @@ package org.seqdoop.hadoop_bam;
 import hbparquet.hadoop.util.ContextUtil;
 import htsjdk.samtools.cram.build.CramContainerIterator;
 import htsjdk.samtools.seekablestream.SeekableStream;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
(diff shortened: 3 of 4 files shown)


akiezun:
@tomwhite can you have a look at this?
@AndreSchumacher what;s your policy regarding java versions?

tomwhite:
@akiezun is this still needed now that htsjdk was fixed to be backwards compatible? Ref: https://github.com/samtools/htsjdk/commit/576fd5b9720712f5cef637918ec8238e30c0347d

akiezun:
The part that is still needed is the passing of validation stringency
through the property I think. Otherwise it will always default to STRICT

On Tuesday, December 8, 2015, Tom White <notifications@github.com> wrote:

> @akiezun <https://github.com/akiezun> is this still needed now that
> htsjdk was fixed to be backwards compatible? Ref: samtools/htsjdk@576fd5b
> <https://github.com/samtools/htsjdk/commit/576fd5b9720712f5cef637918ec8238e30c0347d>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/HadoopGenomics/Hadoop-BAM/pull/40#issuecomment-162843420>
> .
>


-- 
Sent from Gmail Mobile

tomwhite:
Agreed. Worth decoupling from the Java version change.

akiezun:
Will do

On Tuesday, December 8, 2015, Tom White <notifications@github.com> wrote:

> Agreed. Worth decoupling from the Java version change.
> On 8 Dec 2015 12:41, "Adam Kiezun" <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
> > The part that is still needed is the passing of validation stringency
> > through the property I think. Otherwise it will always default to STRICT
> >
> > On Tuesday, December 8, 2015, Tom White <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
> >
> > > @akiezun <https://github.com/akiezun> is this still needed now that
> > > htsjdk was fixed to be backwards compatible? Ref:
> samtools/htsjdk@576fd5b
> > > <
> >
> https://github.com/samtools/htsjdk/commit/576fd5b9720712f5cef637918ec8238e30c0347d
> > >
> > >
> > > —
> > > Reply to this email directly or view it on GitHub
> > > <
> >
> https://github.com/HadoopGenomics/Hadoop-BAM/pull/40#issuecomment-162843420
> > >
> > > .
> > >
> >
> >
> > --
> > Sent from Gmail Mobile
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/HadoopGenomics/Hadoop-BAM/pull/40#issuecomment-162868682
> >
> > .
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/HadoopGenomics/Hadoop-BAM/pull/40#issuecomment-162868975>
> .
>


-- 
Sent from Gmail Mobile

akiezun:
hmm, to use the constructor to CRAMIterator i think i need htsjdk at least at 2.0.0. And to use 2.0.0 I need Java8. Am I missing something?

tomwhite:
You should be able to use the old constructor and call
setValidationStringency().

AndreSchumacher:
Yeah I don't think we want to end support for Java 6 and 7 at the same time, so if you can come up with a different solution that would be great. I imagine Java 6 could go but I think 7 is still widely used. I understand that with the commit that Tom is referring to we can delay the move to htsjdk 2.0.0?

tomwhite:
I opened https://github.com/HadoopGenomics/Hadoop-BAM/pull/41 for the validation stringency part.

PR #42, opened 2015-12-14
Extend SplittingBAMIndexer to allow index creation while writing

tomwhite:
a BAM file, like BAMIndexer in htsjdk.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndex.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndex.java
@@ -80,7 +80,24 @@ public final class SplittingBAMIndex {
 
 	private long   first() { return virtualOffsets.first(); }
 	private long    last() { return prevAlignment(bamSize() - 1); }
-	private long bamSize() { return virtualOffsets.last() >>> 16; }
+	long bamSize() { return virtualOffsets.last() >>> 16; }
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) return true;
+		if (o == null || getClass() != o.getClass()) return false;
+
+		SplittingBAMIndex that = (SplittingBAMIndex) o;
+
+		return virtualOffsets != null ? virtualOffsets.equals(that.virtualOffsets) : that
+				.virtualOffsets == null;
+
+	}
+
+	@Override
+	public int hashCode() {
+		return virtualOffsets != null ? virtualOffsets.hashCode() : 0;
+	}
 
 	/** Writes some statistics about each splitting BAM index file given as an
 	 * argument.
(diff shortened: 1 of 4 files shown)


src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndex.java
```diff
@@ -80,7 +80,24 @@ public Long nextAlignment(final long filePos) {
...
+	long bamSize() { return virtualOffsets.last() >>> 16; }
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) return true;
+		if (o == null || getClass() != o.getClass()) return false;
+
+		SplittingBAMIndex that = (SplittingBAMIndex) o;
+
+		return !(virtualOffsets != null ? !virtualOffsets.equals(that.virtualOffsets) : that
```
AndreSchumacher: How about simplifying this? This should be the same as

```java
return virtualOffsets != null ? virtualOffsets.equals(that.virtualOffsets) : that.virtualOffsets == null);
```

src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndexer.java
```diff
@@ -133,27 +148,104 @@ public static void run(final Configuration conf) throws IOException {
...
+			SAMFileSource fileSource = rec.getFileSource();
+			SAMFileSpan filePointer = fileSource.getFilePointer();
+			writeVirtualOffset(getPos(filePointer));
+		}
+		count++;
+	}
+
+	private long getPos(SAMFileSpan filePointer) {
+		// Use reflection since BAMFileSpan is package private in htsjdk 1.141. Note that
+		// Hadoop-BAM cannot use a later version of htsjdk since it requires Java 8.
```
AndreSchumacher: We should add this to a list of fixes to be done after the upgrade.

AndreSchumacher:
@tomwhite added minor comments, otherwise looks good to me. I guess the idea is to write the index transparently when merging the BAM parts in ReadsSparkSink?

src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndex.java
tomwhite: I think this was auto-generated. Fixed now.

tomwhite:
@AndreSchumacher the main driver for this was to make it possible to generate standard (.bai) indexes and splitting indexes at the same time. BuildBamIndex in GATK could be extended to do this for example.

I'm thinking it may also be useful to generate splitting indexes from BAMRecordWriter too, but this would be a follow-on issue.

Issue #31, opened 2015-11-18
CRAMInputFormat should read reference from HDFS and S3


tomwhite:
CRAMInputFormat, introduced in https://github.com/HadoopGenomics/Hadoop-BAM/pull/28, can currently only read the reference from a local file. We should extend it to support reading reference files (FASTA) from HDFS using an approach like https://github.com/samtools/htsjdk/pull/308 or similar.

tomwhite:
Fixed by #47.

PR #44, opened 2015-12-18
Porting over IndexedBAMInputFormat code.

### LINKED ISSUE #34: Support reading a subset of a BAM
If the Hadoop-BAM index is present, then it should be possible efficiency only load a subset of the BAM, say bases 1-2M of chromosome 5.

This will likely be picked up by the GATK team.

fnothaft:
Resolves #34.

There is some pretty stinky code in here due to class protection issues. Some of these issues are resolved by newer versions of HTSJDK (it seems like the 2.* stream of releases make BAMFileSpan public) or by open PRs (see https://github.com/samtools/htsjdk/pull/413).

Also, this code needs more testing. This is a port of the [Scala version from ADAM](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/java/org/bdgenomics/adam/io/IndexedBamInputFormat.scala), over to Java, but I haven't pulled in unit tests yet. I will try to make a pass back within the next week to get the unit tests in. In the meanwhile, please review this code.

### DIFF
--- a/src/main/java/htsjdk/samtools/PublicBAMFileSpan.java
+++ b/src/main/java/htsjdk/samtools/PublicBAMFileSpan.java
@@ -1,0 +1,38 @@
+// Copyright (c) 2015 Regents of the University of California
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+
+package htsjdk.samtools;
+
+import htsjdk.samtools.Chunk;
+import java.util.List;
+
+/**
+   NOTE!!!!
+   This is really smelly code!!!!
+   
+   The BAMFileSpan and getChunks method are not public in HTSJDK 1.141.0.
+   Newer releases of HTSJDK make this class and method public.
+ */
+public class PublicBAMFileSpan {
+    
+    public static List<Chunk> getChunks(SAMFileSpan sfs) {
+        return ((BAMFileSpan) sfs).getChunks();
+    }
+}
--- a/src/main/java/htsjdk/samtools/PublicDiskBasedBAMFileIndex.java
+++ b/src/main/java/htsjdk/samtools/PublicDiskBasedBAMFileIndex.java
@@ -1,0 +1,40 @@
+// Copyright (c) 2015 Regents of the University of California
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+
+package htsjdk.samtools;
+
+import htsjdk.samtools.seekablestream.SeekableStream;
+
+/**
+   NOTE!!!!
+   This is really smelly code!!!!
+   
+   This class is not public in HTSJDK. We have opened a PR
+   to make the class public upstream. Once that is merged and
+   a new release of HTSJDK is cut, this file should be removed.
+ */
+public class PublicDiskBasedBAMFileIndex
+    extends DiskBasedBAMFileIndex {
+
+    public PublicDiskBasedBAMFileIndex(final SeekableStream stream,
+                                final SAMSequenceDictionary dictionary) {
+        super(stream, dictionary);
+    }
+}
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMFilteredRecordReader.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMFilteredRecordReader.java
@@ -1,0 +1,151 @@
+// Copyright (c) 2015 Regents of the University of California
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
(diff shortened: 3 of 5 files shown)


AndreSchumacher:
@fnothaft thanks a lot for the Java port, I'm hoping to look into this during the next few days. The test failures seem unrelated, very strange.

fnothaft:
No problem! Thanks for looking into this over the busy holiday season, @AndreSchumacher!

src/main/java/org/seqdoop/hadoop_bam/BAMFilteredRecordReader.java
```diff
@@ -0,0 +1,151 @@
...
+    @Override public void initialize(InputSplit spl,
+                                     TaskAttemptContext ctx) 
+        throws IOException {
+        // Check to ensure this method is only be called once (see Hadoop API)
+        if (isInitialized) {
+            close();
+        }
+        isInitialized = true;
+
+        Configuration conf = ContextUtil.getConfiguration(ctx);
```
AndreSchumacher: minor: we could add a bunch of final's here.. as in the parent class.

src/main/java/org/seqdoop/hadoop_bam/BAMFilteredRecordReader.java
```diff
@@ -0,0 +1,151 @@
...
+        FileVirtualSplit split = (FileVirtualSplit) spl;
+        Path file = split.getPath();
+        FileSystem fs = file.getFileSystem(conf);
+
+        this.stringency = SAMHeaderReader.getValidationStringency(conf);
+
+        FSDataInputStream in = fs.open(file);
```
AndreSchumacher: .. and the same here

src/main/java/org/seqdoop/hadoop_bam/IndexedBAMInputFormat.java
```diff
@@ -0,0 +1,116 @@
...
+            // Use index to get the chunks for a specific region, then use them to create InputSplits
+            PublicDiskBasedBAMFileIndex idx = new PublicDiskBasedBAMFileIndex(seekableIdxFile,
+                                                                              dict);
+            int referenceIndex = dict.getSequenceIndex(viewContig);
+
+            // Get the chunks in the region we want (chunks give start and end file pointers into a BAM file)
+            SAMFileSpan sfs = idx.getSpanOverlapping(referenceIndex, viewStart, viewEnd);
+            List<Chunk> regions = PublicBAMFileSpan.getChunks(sfs);
+            Iterator<Chunk> ri = regions.iterator();
+            List<InputSplit> splits = new LinkedList<InputSplit>();
```
AndreSchumacher: Instead of creating the list of input splits from scratch, do you think it would be possibly to use the parent's splits (or maybe rather the ones coming from FileInputFormat) and filter these using the overlapping spans? Or do you think this would not make any sense here?

src/main/java/htsjdk/samtools/PublicBAMFileSpan.java
```diff
@@ -0,0 +1,38 @@
...
+
+import htsjdk.samtools.Chunk;
+import java.util.List;
+
+/**
+   NOTE!!!!
+   This is really smelly code!!!!
+   
+   The BAMFileSpan and getChunks method are not public in HTSJDK 1.141.0.
+   Newer releases of HTSJDK make this class and method public.
```
AndreSchumacher: Is this update blocked by the "Java 8" problem?

src/main/java/htsjdk/samtools/PublicDiskBasedBAMFileIndex.java
```diff
@@ -0,0 +1,40 @@
...
+
+import htsjdk.samtools.seekablestream.SeekableStream;
+
+/**
+   NOTE!!!!
+   This is really smelly code!!!!
+   
+   This class is not public in HTSJDK. We have opened a PR
+   to make the class public upstream. Once that is merged and
+   a new release of HTSJDK is cut, this file should be removed.
```
AndreSchumacher: I suppose making this available as public would also help in resolving https://github.com/HadoopGenomics/Hadoop-BAM/issues/31.

AndreSchumacher:
@fnothaft : I added a few comments above. Thanks for taking the time to go the Scala -> Java route, I feel your pain :)

As commented above: I am wondering whether breaking the pattern of re-using the FileInputFormat splits could somehow could lead to problems down the road, say a large BAM file with a very "permissive" filter leading to too many, too fine splits. Do you think that could be an issue? Also this way does not seem to cover the case of the input splits spanning multiple input files (e.g., via globbing).

Also wanted to point out that here you take a different approach for reading the index. When @tomwhite implemented CRAM there was some discussion about how to read the index (see also https://github.com/HadoopGenomics/Hadoop-BAM/issues/31). I much rather would like to go the way it's done here also in CRAM if possible.

Finally: were you planning to port the tests as well?

fnothaft:
Thanks for the review @AndreSchumacher; I'm offline this week, but will take a detailed look over the comments when I'm back online. I'll also compare against #31 to see @tomwhite's approach.

> As commented above: I am wondering whether breaking the pattern of re-using the FileInputFormat splits could somehow could lead to problems down the road, say a large BAM file with a very "permissive" filter leading to too many, too fine splits. Do you think that could be an issue? Also this way does not seem to cover the case of the input splits spanning multiple input files (e.g., via globbing).

Yes, both of these are valid concerns. Again, this code is just a direct port of the code that we wrote in ADAM for doing indexed loads where the assumption was that we were doing small range queries (e.g., <50kbp) against a single file. That was a reasonable simplifying assumption for what we were doing, but isn't necessarily a realistic assumption for general workloads.

> Finally: were you planning to port the tests as well?

Yes, I'll try to get those next week.

src/main/java/org/seqdoop/hadoop_bam/BAMFilteredRecordReader.java
```diff
@@ -0,0 +1,151 @@
...
+          if (r == null) {
+              return false;
+          } else {
+              int start = r.getStart();
+              int end = r.getEnd();
+
+              if ((r.getContig() == viewContig) &&
+                  (((start >= viewStart) && (end <= viewEnd))
+                   || ((start <= viewStart) && (end >= viewStart) && (end <= viewEnd))
+                   || ((end >= viewEnd) && (start >= viewStart) && (start <= viewEnd)))) {
```
tomwhite: I think this condition is overly restrictive, as it doesn't cover the case where the interval (viewStart, viewEnd) is wholly contained in (start, end). I tried changing the condition to `htsjdk.samtools.util.CoordMath.overlaps(start, end, viewStart, viewEnd)` and it worked for the testing I've been doing for https://github.com/broadinstitute/gatk/issues/1399.

Unmapped reads with a position need to be taken care of too - see https://github.com/broadinstitute/gatk/blob/master/src/main/java/org/broadinstitute/hellbender/engine/spark/datasources/ReadsSparkSource.java#L210

src/main/java/org/seqdoop/hadoop_bam/IndexedBAMInputFormat.java
```diff
@@ -0,0 +1,116 @@
...
+    extends BAMInputFormat {
+
+    private static Path filePath;
+    private static Path indexFilePath;
+    private static String viewContig;
+    private static int viewStart;
+    private static int viewEnd;
+    private static SAMSequenceDictionary dict;
+
+    public static void setVars(Path bamPath,
```
tomwhite: Hadoop-BAM tends to use configuration or instance methods for passing information to the input or output format, rather than statics. @AndreSchumacher - do you have a preference?

Also, it would be good to think about how multiple intervals could be specified (even if it's not implemented in this PR). Perhaps use htsjdk's QueryInterval?

src/main/java/org/seqdoop/hadoop_bam/IndexedBAMInputFormat.java
```diff
@@ -0,0 +1,116 @@
...
+                               Path baiPath,
+                               String contig,
+                               int start,
+                               int end,
+                               SAMSequenceDictionary sd) {
```
tomwhite: Could the dictionary be optional, and retrieved from the BAM if not specified?

src/main/java/org/seqdoop/hadoop_bam/IndexedBAMInputFormat.java
tomwhite: Another case to think about (possibly as a follow on) is what if the intervals are very large - much larger than an HDFS block (e.g. a whole chromosome). Then we'd want the to break up the split roughly on HDFS block boundaries, just like  BAMInputFormat does already.

src/main/java/org/seqdoop/hadoop_bam/IndexedBAMInputFormat.java
AndreSchumacher: Good point. About the configuration versus instance methods: I have a slight preference towards configuration-based because IMHO it makes debugging easier (all the settings in one place, easy to search in Spark UI, etc.). We could static helper methods that add settings to the config if required, e.g.,

``public static void setBaiPath(Configuration configuration, String path) {
    if (configuration != null) configuration.set(IndexedBAMInputFormat.BAM_INDEX_PATH, path);
}
``

tomwhite:
@fnothaft would you like me to continue the work on this one? I also just realized there may be some interaction with #49, so we might need to figure out how to support both at the same time.

tomwhite:
I took the liberty of working on this, see https://github.com/tomwhite/Hadoop-BAM/compare/paired-reads-in-same-split...tomwhite:use-bai. It's not yet complete, but this major changes are:
* based on logic from https://github.com/HadoopGenomics/Hadoop-BAM/pull/49
* the feature is now built into BAMInputFormat/BAMRecordReader
* filters existing splits rather than creating new splits based on chunks
* supports multiple intervals
* includes a test

AndreSchumacher:
@tomwhite +1 on the filtering-existing-splits approach which also makes treatment of several intervals a lot easier. If @fnothaft does not mind could you maybe submit this as an PR (dependent on #49)?

tomwhite:
We should be able to close this now that #59 is done.

AndreSchumacher:
True. Thanks again @fnothaft for porting this over from ADAM. I'm closing this now.

Issue #46, opened 2016-01-02
move to Java 8


AndreSchumacher:
Since HTSJDK now requires Java 8 we need to ensure compatibility with 1.8 as well. I don't think we want to maintain versions to the easiest would be just to upgrade.

Based on the discussion here https://github.com/samtools/htsjdk/issues/180 it is safe to say the issue will cause controversy.

edit: We should probably start off with working on a separate branch and testing that. Also improving the test coverage would not hurt.

droazen:
Has the proposal to move to Java 8 caused any actual controversy yet? There is at least one branch already in existence that switches the project to Java 8 and passes tests (https://github.com/HadoopGenomics/Hadoop-BAM/pull/47).

fnothaft:
I am OK with a move to Java 8. CC @massie @heuermh @laserson @ryan-williams for further comment.

heuermh:
We support JDK8 but do not require it, I would prefer our dependencies do the same.  If htsjdk is forcing our hand then there's not much we can do.  Nice of them to close comments on that issue and make decisions in private that affect all downstream projects.

droazen:
@heuermh This decision was hardly made in private. There was an announcement last March on the Samtools-devel list that support for Java 6 and 7 in htsjdk would end in October 2015 (6 months after the official end of Oracle's public updates for Java 7). This announcement was followed by a one-month public comment period that took place here: https://github.com/samtools/htsjdk/issues/209. The resulting Java Minimum Version Support Policy has been posted on htsjdk's github README for almost a year now...

ilveroluca:
In my opinion the fact that they're not supporting slightly older Java releases is still a pain, regardless of how long ago they announced it.  The time we spend chasing Java isn't spent on fixing issues or adding functionality that matters more :-(  Nevertheless, if we need it I guess we have no choice, and neither will our dependencies.

droazen:
That's a fair complaint -- the htsjdk policy was motivated primarily by Oracle stopping support for Java 7 in April 2015, though (http://www.oracle.com/technetwork/java/eol-135779.html).

AndreSchumacher:
Yeah I guess the only sensible thing to do is to require Java 8 as well. Let's make one more release before that and then do the move.

fnothaft:
I think that is a reasonable approach. It sounds like the next release of Spark (2.0.0) may drop Java 7 and Scala 2.10 anyways, which would render Java 7 compatibility moot from our side.

PR #47, opened 2016-01-05
Use jsr203-hadoop library to read CRAM reference from HDFS

tomwhite:
Note that this needs Java 8, so depends on https://github.com/HadoopGenomics/Hadoop-BAM/issues/46

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
 
     <properties>
         <java.version>1.8</java.version>
-        <htsjdk.version>1.141</htsjdk.version>
+        <htsjdk.version>2.0.1</htsjdk.version>
         <!-- CHANGE THIS FOR A DIFFERENT VERSION OF HADOOP -->
         <hadoop.version>2.2.0</hadoop.version>
         <!--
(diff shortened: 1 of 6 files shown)


tomwhite:
This addresses https://github.com/HadoopGenomics/Hadoop-BAM/issues/31

damiencarol:
@tomwhite I don't understand why Java 8 is needed. :/  It's because of _jsr203-hadoop_ ?
If so, I can force Java 7 in _jsr203-hadoop_

tomwhite:
@damiencarol it's because this change relies on a newer version of htsjdk, which now requires Java 8. (I agree that jsr203-hadoop only requires Java 7.)

pom.xml
```diff
@@ -277,6 +282,36 @@
             <version>1.10.19</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
```
AndreSchumacher: This artifact does not seem to exist to versions 1.1.2 and 1.2.1 (currently still listed as supported). I guess it's OK since this is for the tests only but is there a way to make it work for these as well? I'm not saying we should replicate the test support classes.

tomwhite: You're right that the artifact does not exist in Hadoop 1. The API for the the mini cluster is different in that release too, so it would be tricky to support both. This is for testing, so it may be possible to exclude the test somehow, although it probably need a special Maven profile.

tomwhite:
In the interests of moving this forward, to address the Hadoop 1 issue I've added a profile so that the Hadoop 2 test dependencies (and the CRAM unit tests) can be excluded by adding `-Dhadoop1` to the command line.

AndreSchumacher:
Thanks @tomwhite! I think this is ready to be merged now.

Issue #43, opened 2015-12-17
Add an option to ensure paired reads are in the same split


tomwhite:
If a BAM has consecutive paired reads then it is sometimes desirable for them to be in the same split - for example Mark Duplicates could take advantage of this to avoid an initial sort of the BAM.

The BAMRecordReader could ensure that paired reads are always in the same split by starting with the first record in the split that is unpaired or the first of a pair. I.e. it would only return the first record if the following expression was true:

```!record.getReadPairedFlag() || record.getFirstOfPairFlag()```

Similarly it would return an extra record at the end of the split if the last record was paired and was the first of a pair:

```record.getReadPairedFlag() && record.getFirstOfPairFlag()```

PR #49, opened 2016-01-07
Add an option to ensure paired reads are in the same split.

tomwhite:
Here's my first attempt at this (#43). Probably needs more testing beyond the unit tests I've written.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
@@ -53,6 +53,16 @@ public class BAMInputFormat
 	// set this to true for debug output
 	public final static boolean DEBUG_BAM_SPLITTER = false;
 
+	/**
+	 * If set to <code>true</code>, ensure that for paired reads both reads in a pair are
+	 * always in the same split for queryname-sorted BAM files.
+	 * <p>
+	 * Note: only use this option if for all paired reads both reads in each pair are
+	 * present, otherwise it is possible that reads may be silently dropped.
+	 */
+	public static final String KEEP_PAIRED_READS_TOGETHER_PROPERTY =
+			"hadoopbam.bam.keep-paired-reads-together";
+
 	private Path getIdxPath(Path path) {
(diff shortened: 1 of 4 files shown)


droazen:
Would someone more familiar with this codebase than us (the GATK team) be available to take a look at this change and let us know if it's sane? Our goal is to be able to safely accept queryname-sorted input in our Spark tools without having to do a `groupByKey` on the read name and without having to worry about read pairs getting assigned to different partitions. Thanks!

droazen:
@laserson, perhaps?

laserson:
I'm not really familiar with this codebase.  I'll tag @fnothaft or @ryan-williams

droazen:
@AndreSchumacher would you be able to take a look at this one when you have time?

AndreSchumacher:
From my understanding the logic makes sense (assuming that queryname sort order does indeed place both reads of a pair consecutively in the right order which seems to be the case).

The case I was wondering about: let's say in some QC step you filter out reads disregarding of whether they are paired or not and later on you want to read the output of that back in, using this option. If there is a second read of a pair at the beginning of some split whose mate was filtered out then this read will be silently dropped. Is it enough to have a warning in the docs that if one uses this option one needs to ensure that all read pairs are properly paired?

tomwhite:
Thanks for taking a look @AndreSchumacher. I think your analysis is right, although I don't know how often that would come up in practice. Perhaps @droazen could comment? I've added some documentation to the property to warn about this possibility.

tomwhite:
I've updated the code to simplify the logic a little, so that #34 is easier to integrate. The idea is to change `BamRecordReader#nextKeyValue` so that it is a while loop rather than a single if statement, since a while loop allows reads to be filtered out (if they are not the first in a pair for this issue, or if they aren't in an interval for #34).

AndreSchumacher:
Thanks @tomwhite for simplifying the logic, I think it's now much more clear. One final comment: I don't think it could cause problems but I would feel better also initializing <code>readPair</code> and <code>lastOfPair</code> in the <code>initialize()</code> function.

tomwhite:
Thanks for the review @AndreSchumacher. I've added the initialization you suggested, squashed the commits, and will now merge.

PR #50, opened 2016-01-15
Pass a reference source through to SAMHeaderReader.readSAMHeaderFrom.

cmnbroad:
readSAMHeaderFrom will fail on CRAM files using newer versions of htsjdk ((2.0.1) because a reference source is required for CRAM files, so we need to pass one through.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/util/SAMHeaderReader.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/util/SAMHeaderReader.java
@@ -24,7 +24,10 @@ package org.seqdoop.hadoop_bam.util;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
+import java.nio.file.Paths;
 
+import htsjdk.samtools.cram.ref.ReferenceSource;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/seqdoop/hadoop_bam/util/SAMHeaderReader.java
```diff
@@ -70,4 +78,19 @@ public static ValidationStringency getValidationStringency(
...
 		return p == null ? null : ValidationStringency.valueOf(p);
 	}
+
+	public static ReferenceSource getReferenceSource(
+			final Configuration conf)
+	{
+		//TODO: There isn't anything particularly CRAM-specific about reference source or validation
+		// stringency other than that a reference source is required for CRAM files. We should move
+		// the reference source and validation stringency property names and utility methods out of
+		// CRAMInputFormat and SAMHeaderReader and combine them together into a single class for extracting
```
AndreSchumacher: Agree. Maybe something akin to HTSJDK's ReferenceSource? But for now I think this works fine.

AndreSchumacher:
Thanks for the PR. A general question: could we hold off with this until https://github.com/HadoopGenomics/Hadoop-BAM/pull/47 is merged? Then we would not propagate the dependence of having the reference on the local filesystem.

cmnbroad:
Sure. I'll update the PR to reflect that.

cmnbroad:
Ok, I updated this per the comments above, but its dependent on #47 now (actually, its only dependent on the java and htsjdk upgrades, which I didn't reproduce in this PR). So both the compile and  tests here will fail until #47 is merged.

cmnbroad:
Once #47 is merged in I'll rebase and push these again.

AndreSchumacher:
@cmnbroad now that #47 is merged, could you rebase this PR? Thanks!

cmnbroad:
Done.

PR #52, opened 2016-01-19
Run findbugs on main package.

tomwhite:


tomwhite:
The Java 6 build is failing since findbugs is using Java 7. This can wait until we remove Java 6 support.

src/main/java/org/seqdoop/hadoop_bam/BCFSplitGuesser.java
```diff
@@ -399,18 +400,18 @@ private int guessNextBCFPos(long cpVirt, int up, int cSize) {
...
 
 				cinSeek(cpVirt | up+24);
-				cin.read(buf.array(), 0, 4);
+				IOUtils.readFully(cin, buf.array(), 0, 4);
 				final int alleleInfo = buf.getInt(0);
 
 				final int alleleCount = alleleInfo >> 16;
 				final int infoCount   = alleleInfo & 0xffff;
-				if (alleleCount < 0 || infoCount < 0)
+				if (alleleCount < 0) // don't check infoCount since it is always nonnegative
```
AndreSchumacher: Good point!

src/main/java/org/seqdoop/hadoop_bam/VCFRecordReader.java
```diff
@@ -84,11 +84,11 @@
 		reader = new AsciiLineReader(ins);
         it = new AsciiLineReaderIterator(reader);
 		
-		final Object h = codec.readHeader(it);
-		if (!(h instanceof FeatureCodecHeader) || !(((FeatureCodecHeader)h).getHeaderValue() instanceof VCFHeader))
+		final FeatureCodecHeader h = codec.readHeader(it);
```
AndreSchumacher: I wonder.. any idea what was the original reason for having Object there? Was that required by an old HTSJDK API?

AndreSchumacher:
These are plenty of smaller changes. On a quick glance they all seem to make sense. I quickly checked and it seems that `org.apache.hadoop.io.IOUtils` at least seems to have existed in Hadoop 1 as well. So I don't see any reason not to do those fixes. Did you find them all with the tool?

src/main/java/org/seqdoop/hadoop_bam/VCFRecordReader.java
tomwhite: Possibly - I don't know. We mandate a minimum version, don't we?

tomwhite:
Yes, all the changes were found with FindBugs.

PR #55, opened 2016-01-22
Move to Java 8. Fixes #46.

### LINKED ISSUE #46: move to Java 8

tomwhite:
This is a standalone PR to move to Java 8. There are a few other outstanding PRs that depend on this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: java
 jdk:
   - oraclejdk8
-  - oraclejdk7
-  - openjdk7
-  - openjdk6
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


akiezun:
+1 from GATK

AndreSchumacher:
Thanks @tomwhite

PR #56, opened 2016-01-25
Remove unused copies of MR library classes. The classes had also

tomwhite:
been repackaged, so they should not have been used outside Hadoop-BAM.


tomwhite:
Do you think we could merge this @AndreSchumacher? It should be uncontroversial (unless I'm missing something).

AndreSchumacher:
I believe that older Hadoop versions (0.20.X) may still need this patched version of MultipleOutputs. I had a quick look and it only seems to be used in one of the CLI plugins. So my proposed fix is to take away the _Older stable versions as far back as 0.20.2 should also work_ away from README.md and then we can merge this.

tomwhite:
Thanks for the explanation @AndreSchumacher. The CLI plugins use the `MultipleOutputs` class from the `org.apache.hadoop.mapreduce.lib.output` package, not from `hadoop.org.apache.hadoop.mapreduce.lib.output` (note the leading `hadoop` part). And the build doesn't seem to do any relocation of the latter package. Which leads me to conclude the versions in the `hadoop` package are not used by Hadoop-BAM.

I've removed the note from the README as you suggested, so we can merge this.

AndreSchumacher:
Thanks @tomwhite for the correction, I had missed that. I'll merge this now.

PR #59, opened 2016-02-01
Support filtering by interval in BAMInputFormat using a bai index file

tomwhite:
This is an continuation of the work in #44 by @fnothaft, but with the following changes:

* the feature is now built into BAMInputFormat/BAMRecordReader
* filters existing splits, then trims the size using chunk boundaries
* supports multiple intervals
* includes tests

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
@@ -22,6 +22,21 @@
 
 package org.seqdoop.hadoop_bam;
 
+import htsjdk.samtools.BAMFileSpan;
+import htsjdk.samtools.BAMIndex;
+import htsjdk.samtools.Chunk;
+import htsjdk.samtools.DiskBasedBAMFileIndex;
+import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMSequenceDictionary;
+import htsjdk.samtools.util.CoordMath;
+import htsjdk.samtools.util.Interval;
+import htsjdk.samtools.util.Locatable;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FileSystem;
+import org.seqdoop.hadoop_bam.util.SAMHeaderReader;
 import org.seqdoop.hadoop_bam.util.WrapSeekable;
 import hbparquet.hadoop.util.ContextUtil;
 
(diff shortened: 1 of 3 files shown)


src/main/java/htsjdk/samtools/PublicDiskBasedBAMFileIndex.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ NOTE!!!!
+ This is really smelly code!!!!
+
+ This class is not public in HTSJDK. We have opened a PR
+ to make the class public upstream. Once that is merged and
+ a new release of HTSJDK is cut, this file should be removed.
+ */
+public class PublicDiskBasedBAMFileIndex extends DiskBasedBAMFileIndex {
```
AndreSchumacher: Looking at https://github.com/samtools/htsjdk/commit/23eb1f02090a507ba3510f20d8b23e94a9e47516 it seems the base class is now public so maybe this is not needed anymore?

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -235,6 +284,81 @@ private int addProbabilisticSplits(
...
 
+	private List<InputSplit> filterByInterval(List<InputSplit> splits, Configuration conf)
+			throws IOException {
+		List<Locatable> intervals = getIntervals(conf);
+		if (intervals == null) {
+			return splits;
+		}
+
+		// Get the chunks in the intervals we want (chunks give start and end file pointers
+		// into a BAM file) by looking in all the indexes for the BAM fils
```
AndreSchumacher: Small typo: 'BAM files'

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -235,6 +284,81 @@ private int addProbabilisticSplits(
...
+			}
+			if (overlaps) {
+				filteredSplits.add(new FileVirtualSplit(virtualSplit.getPath(), newStart, newEnd,
+						virtualSplit.getLocations()));
+			}
+		}
+		return filteredSplits;
+	}
+
+	private static boolean overlaps(long start, long end, long start2, long end2) {
```
AndreSchumacher: Should we use <code>CoordMath.overlaps(..)</code> instead here as well as below?

src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
```diff
@@ -207,10 +213,36 @@ public static long getKey0(int refIdx, int alignmentStart0) {
 				}
 			}
 
+			if (!overlaps(r, intervals)) {
```
AndreSchumacher: We should probably keep an eye on this line whether it becomes a bottleneck when we have many intervals to filter from and then possibly switch data structures for the intervals. But let's leave it as is for now.

src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
```diff
@@ -207,10 +213,36 @@ public static long getKey0(int refIdx, int alignmentStart0) {
...
+
 			key.set(getKey(r));
 			record.set(r);
 			return true;
 		}
 		return false;
 	}
+
+	private static boolean overlaps(SAMRecord r, List<Locatable> intervals) {
+		if (intervals == null) {
```
AndreSchumacher: Maybe switch the condition to <pre>intervals == null || (r.getReadUnmappedFlag() && r.getAlignmentStart() == SAMRecord.NO_ALIGNMENT_START)</pre>? Then the check below could be slightly simplified but the benefit would be to completely skip interval checking for unmapped reads.

AndreSchumacher:
@tomwhite thanks for the PR! I added a few minor comments above otherwise this looks good to me.

src/main/java/htsjdk/samtools/PublicDiskBasedBAMFileIndex.java
tomwhite: Yes, a new version of htsjdk has been released now so I can remove this.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
tomwhite: Fixed.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
tomwhite: I wanted to, but it takes ints not longs, hence this method.

src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
tomwhite: OK.

src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
tomwhite: Good point. Updated.

tomwhite:
Addressed feedback and squashed the commits. I appreciate the review @AndreSchumacher.

AndreSchumacher:
Thanks @tomwhite! Merging this now.

Issue #63, opened 2016-02-08
CRAM Writing doesn't properly handle the case where headers are written to each part


cmnbroad:
I discovered this when testing GATK with sharded cram output. I'll have a fix shortly.

PR #64, opened 2016-02-08
Fix for CRAM header writing (issue #63).

cmnbroad:
Fixes https://github.com/HadoopGenomics/Hadoop-BAM/issues/63. CRAM writing worked when the headers were not included in the output, but didn't correctly handle the case where they were. CRAMOutputFormat and CRAMRecordWriter are now symmetric with their BAM counterparts.


tomwhite:
+1

Issue #68, opened 2016-02-17
block gzipped VCF not recognized as VCF


akiezun:
I have a block gzipped vcf and I want to read it using a call like this:
```
ctx.newAPIHadoopFile(
                vcf, VCFInputFormat.class, LongWritable.class, VariantContextWritable.class,
                new Configuration());
```

`VCFFormat.inferFromData` falsely thinks it's a BCF because it's gzipped and starts with a `1f` byte. And so my code breaks. The vcf is attached. Block gzipped VCF should be recognized as VCFs not BCFs. 

[count_variants.blockgz.gz](https://files.zenhub.io/56c4e45d5d3a5c4e4bb3ccf5)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

tomwhite:
@akiezun I created a fix in #70. Could you test to see if it works for you?

akiezun:
thanks @tomwhite - does it work for blocked gzipped files? (we typically use bgzip from samtools). http://www.htslib.org/doc/tabix.html 
You can use the file I attached: count_variants.blockgz.gz

tomwhite:
I posted some new code in #70. I tested it with count_variants.blockgz.gz and it could read the file fine. The PR has more tests, including a large file with multiple splits.

tomwhite:
Fixed in #70.

Issue #23, opened 2015-10-13
Is gz compressed VCF file supported?


tkrishp:
Can someone clarify if vcf.gz is supported by VCFInputFormat?

AndreSchumacher:
There is support for BCF which is a compressed version of VCF (similarly to BAM and SAM). See [here for a description of the format](http://www.1000genomes.org/wiki/analysis/variant-call-format/bcf-binary-vcf-version-2 ). Also most Hadoop versions should handle gzip compressed files more or less transparently but you loose the benefit of file splitting in this case.

tkrishp:
I tried importing vcf.gz but it errors out with 'Invalid BCF format' message.
line # 56 in VCFFormat.java
       `case 0x1f: return BCF;`
This is making the framework treat gz compressed VCF file as BCF, causing the error. Let me know if I'm missing something, thanks.

akiezun:
duplicate of #68 or vice versa - Hadoop-BAM does fail on such files. It's important to support them (VCFCodec will decode them no problem)

@AndreSchumacher can you mark one of those reports are a dup and accept a +1 vote for this feature?

tomwhite:
Fixed in #70

PR #70, opened 2016-02-24
Support gzip compressed VCF files.

tomwhite:
This uses some Hadoop machinery for reading compressed files, although I've only handled the non-splittable compression case (which includes gzip), since I don't think splittable compression (like bzip2) is used much in genomics.

Addresses #23 and #68.


src/main/java/org/seqdoop/hadoop_bam/VCFRecordReader.java
```diff
@@ -126,7 +163,7 @@
 	@Override public VariantContextWritable getCurrentValue() { return vc; }
 
 	@Override public boolean nextKeyValue() throws IOException {
-		if (!it.hasNext() || it.getPosition() >= length)
+		if (!it.hasNext() || getPosition() > length)
```
tomwhite: Note the boundary change here, which was needed to get the test for test.vcf.gz to pass. I think it's needed because the underlying stream buffers, so it reads to the end of the file before finishing the iteration. Note that this condition is now the same as the one above (line 115/152), so it should be OK.

tomwhite:
This is incomplete since it doesn't handle the block gzipped case properly. It will work, since block compressed gzip is still gzip, but the way I've done it here will treat the input as a single split (like regular Hadoop), which is inefficient. This PR needs to use the code for handling block gzipped splits properly in the case of VCF (not just BCF, which VCFInputFormat has done for a long time).

Another slight complication is that it's not possible to determine the compression type (gzip vs bgzf) from the file extension, since .gz may be used for either (e.g. the bgzip tool adds a .gz extension by default). However, it's possible to detect the type by looking at the stream header.

tomwhite:
I've updated this PR with code to support block compressed VCFs. I've added a BGZFCodec class that is a Hadoop compression codec that handles splitting properly. I also changed VCFRecordReader to use Hadoop's LineRecordReader, which has been tested extensively on text file processing (including tests for split boundaries, etc). I extended the test to check that VCF records are not lost, and this is now run against several VCF files, including a large multi-split VCF file.

Note that the code needs access to an internal field in htsjdk's BlockCompressedInputStream, so I copied the code here, as a stopgap until https://github.com/samtools/htsjdk/pull/522 is resolved. 

This is now ready for review.

tomwhite:
@AndreSchumacher would you be able to have a look at this when you have time?

src/main/java/htsjdk/samtools/util/ModifiedBlockCompressedInputStream.java
```diff
@@ -0,0 +1,524 @@
...
+import java.io.InputStream;
+import java.io.RandomAccessFile;
+import java.net.URL;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.Arrays;
+
+/*
+ * Modified from htsjdk.samtools.util.BlockCompressedInputStream to add the endOfBlock()
+ * method.
```
cmnbroad: Would we still need this file if we upgraded to htsjdk 2.2.1, which I think includes the endOfBlock PR ?

src/main/java/org/seqdoop/hadoop_bam/VCFFormat.java
```diff
@@ -59,4 +61,9 @@ public static VCFFormat inferFromData(final InputStream in) throws IOException {
 		}
 		return null;
 	}
+
+	public static boolean isBGZF(final InputStream in) throws IOException {
```
cmnbroad: This seems a bit out of place to me, since its not really VCF-specific, and there are other call sites that access BlockCompressedInputStream.isValidFile directly. It seems like we could get rid of this and the corresponding tests, since its just a pass-through. If you do want to keep it though, we should probably change BCFSplitGuesser to call it rather than calling isValidFile directly.

src/main/java/org/seqdoop/hadoop_bam/VCFInputFormat.java
```diff
@@ -149,6 +155,33 @@ public VCFFormat getFormat(final Path path) {
...
+			}
+			if (!splittable) {
+				System.err.printf("Warning: %s is not splittable, consider using block " +
+						"compressed gzip (BGZF).\n", filename);
+			}
+			return splittable;
+		}
+		return codec instanceof SplittableCompressionCodec;
+	}
+
```
cmnbroad: Can/should detect the case where the application hasn't established the BGZF codec property, but provides a BGZF.gz file. I think we'll silently succeed and treat it as non-splittable gzip ?

src/main/java/org/seqdoop/hadoop_bam/util/BGZFSplitGuesser.java
```diff
@@ -23,25 +23,37 @@
...
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.Arrays;
 
 import htsjdk.samtools.util.BlockCompressedInputStream;
 
 import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.Seekable;
 
 public class BGZFSplitGuesser {
```
cmnbroad: Can/should this extend BaseSplitGuesser  ? It seems like duplicated code.

cmnbroad:
A few minor comments/questions. Back to @tomwhite

src/main/java/htsjdk/samtools/util/ModifiedBlockCompressedInputStream.java
tomwhite: Upgraded to htsjdk 2.2.1 and removed this file.

src/main/java/org/seqdoop/hadoop_bam/VCFFormat.java
tomwhite: Agreed that this is not really needed. I've removed the method and call isValidFile directly.

src/main/java/org/seqdoop/hadoop_bam/util/BGZFSplitGuesser.java
tomwhite: The duplication is already present, and is not trivial to remove as it has a slightly different signature and method body. See https://github.com/HadoopGenomics/Hadoop-BAM/pull/77 where I removed some duplication. I'd like to tackle the remaining duplication in a separate issue.

src/main/java/org/seqdoop/hadoop_bam/VCFInputFormat.java
tomwhite: If BGZFCodec is not registered then an application will just use regular gzip, with no warning. Since Hadoop-BAM is usually used as a library from GATK or ADAM, I thought that would be OK. It should be possible to see if GzipCodec is used, and then issue a warning if indeed the file is BGZF. Is that what you mean?

tomwhite:
Many thanks for the review @cmnbroad. I've addressed all your points, except one - where I have a question about the intended behaviour.

src/main/java/org/seqdoop/hadoop_bam/VCFInputFormat.java
cmnbroad: Yeah, thats what I meant. It would just be a shame to fall back to gzip since that won't split right ? Certainly not a requirement, just a nice to have.

src/main/java/org/seqdoop/hadoop_bam/util/BGZFSplitGuesser.java
cmnbroad: Fair enough.

cmnbroad:
LGTM.

src/main/java/org/seqdoop/hadoop_bam/VCFInputFormat.java
tomwhite: I added a warning for this case.

PR #74, opened 2016-03-01
Add a couple of simple unit tests (for SAM and FASTA inputs) to increase code coverage

tomwhite:


tomwhite:
@cmnbroad do you think you could review this when you have time?

cmnbroad:
Sure. The one thing I notice is that there are a few places where the various readers aren't being closed. Otherwise LGTM. Thanks for adding those.

tomwhite:
Thanks for reviewing this @cmnbroad! I added close statements, and will merge once it's passing tests.

PR #77, opened 2016-03-14
Split guesser refactoring and tests

tomwhite:
This removes some, but not all, of the split guessing code for BGZF boundaries. BAMSplitGuesser and BCFSplitGuesser now share code (which is already tested). BGZFSplitGuesser does not, but now has a test, which will improve coverage.


heuermh:
FYI I think this may address an issue we have downstream in ADAM (https://github.com/bigdatagenomics/adam/issues/896); I'll try building from your branch to see if it helps.

tomwhite:
@heuermh thanks for your comment. This PR is a refactor and doesn't change any functionality, so I wouldn't expect it to fix the issue you are seeing. Perhaps you could open a separate Hadoop-BAM issue for that.

PR #80, opened 2016-03-29
Fix case where large BAM header causes "no reads in first split" error.

tomwhite:
This should fix https://github.com/bigdatagenomics/adam/issues/896. I tried the fix on a BAM file supplied by @heuermh and the exception no longer occurs.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java
@@ -22,7 +22,13 @@
 
 package org.seqdoop.hadoop_bam;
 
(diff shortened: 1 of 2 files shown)


heuermh:
Works for me in ADAM.  Thank you, @tomwhite!

tomwhite:
Thanks for trying it out @heuermh. I'd like to commit this soon unless there are any objections.

AndreSchumacher:
That's an interesting bug. Maybe BAM headers have grown in size of the years? :-) Thanks @tomwhite for the fix. This looks good to me.

tomwhite:
Thanks for the review @AndreSchumacher.

Issue #71, opened 2016-02-24
add support for reading intervals of VCF file


akiezun:
when reading a vcf file (raw or bgzipped) we need to be able to read intervals of it. Similarly to using a bai for BAM you could use tribble or tabix indexes to read parts of the file. Support for this would be very valuable for users (eg gatk)

tomwhite:
Fixed in #81.

PR #81, opened 2016-04-05
Support filtering by interval in VCFInputFormat using a tabix index

tomwhite:
This addresses #71. It depends on #70 so that needs to be reviewed and merged first.


tomwhite:
I've updated this to use OverlapDetector too. @akiezun, can you take a look at this one too please?

akiezun:
:1 from me

PR #82, opened 2016-04-14
Use htsjdk OverlapDetector to optimize intervals/read overlap calcula…

tomwhite:
…tion.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
@@ -22,7 +22,11 @@
 
 package org.seqdoop.hadoop_bam;
 
+import htsjdk.samtools.util.Interval;
+import htsjdk.samtools.util.Locatable;
+import htsjdk.samtools.util.OverlapDetector;
 import java.io.IOException;
+import java.util.Collection;
 import java.util.List;
 
 import org.apache.hadoop.conf.Configuration;
@@ -40,8 +44,6 @@ import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMFileHeader.SortOrder;
 import htsjdk.samtools.SAMRecord;
 import htsjdk.samtools.util.BlockCompressedInputStream;
-import htsjdk.samtools.util.CoordMath;
-import htsjdk.samtools.util.Locatable;
 
 import org.seqdoop.hadoop_bam.util.MurmurHash3;
 import org.seqdoop.hadoop_bam.util.SAMHeaderReader;
@@ -65,7 +67,8 @@ public class BAMRecordReader
 	private boolean keepReadPairsTogether;
 	private boolean readPair;
 	private boolean lastOfPair;
-	private List<Locatable> intervals;
+	private List<Interval> intervals;
+	private OverlapDetector<Interval> overlapDetector;
 
 	/** Note: this is the only getKey function that handles unmapped reads
 	 * specially!
@@ -162,6 +165,11 @@ public class BAMRecordReader
 		readPair = false;
 		lastOfPair = false;
 		intervals = BAMInputFormat.getIntervals(conf);
+		if (intervals != null) {
+			overlapDetector = new OverlapDetector<>(0, 0);
+			overlapDetector.addAll(intervals, intervals);
+		}
+
 	}
 	@Override public void close() throws IOException { bci.close(); }
 
@@ -211,7 +219,7 @@ public class BAMRecordReader
 				}
 			}
 
-			if (!overlaps(r, intervals)) {
+			if (!overlaps(r)) {
 				continue;
 			}
 
@@ -222,25 +230,23 @@ public class BAMRecordReader
 		return false;
 	}
 
-	private static boolean overlaps(SAMRecord r, List<Locatable> intervals) {
+	private boolean overlaps(SAMRecord r) {
 		if (intervals == null ||
 				(r.getReadUnmappedFlag() && r.getAlignmentStart() == SAMRecord.NO_ALIGNMENT_START)) {
 			return true;
 		}
-		for (Locatable interval : intervals) {
-			if (r.getReadUnmappedFlag()) {
+		if (r.getReadUnmappedFlag()) { // special case for unmapped reads with coordinate set
+			for (Locatable interval : intervals) {
 				if (interval.getStart() <= r.getStart() && interval.getEnd() >= r.getStart()) {
 					// This follows the behavior of htsjdk's SamReader which states that
 					// "an unmapped read will be returned by this call if it has a coordinate for
 					// the purpose of sorting that is in the query region".
 					return true;
 				}
-			} else if (r.getContig().equals(interval.getContig()) &&
-					CoordMath.overlaps(r.getStart(), r.getEnd(),
-						interval.getStart(), interval.getEnd())) {
-				return true;
 			}
 		}
-		return false;
+		final Interval interval = new Interval(r.getContig(), r.getStart(), r.getEnd());
+		Collection<Interval> overlaps = overlapDetector.getOverlaps(interval);
+		return !overlaps.isEmpty();
 	}
 }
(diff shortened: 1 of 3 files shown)


akiezun:
👍  looks great and works great too

akiezun:
I used it here and got a 10x speedup: https://github.com/broadinstitute/gatk/issues/1532

tomwhite:
Thanks for taking a look @akiezun. I've made an update which handles the case for unmapped reads with the start coordinate set (and included a test), which is handled in the existing code in the same way as htsjdk's SamReader. Such reads are presumably rare so there shouldn't be much of a performance impact.

akiezun:
looks good. @AndreSchumacher can you review this too? We'd like to merge it soon as it has a huge performance benefit

tomwhite:
@cmnbroad could you possibly have a look at this one? Thanks.

ddurnev:
I've tested the patch with big .bed file (6.8 Mb, 240K intervals) -  and for me performance is still not good: only computing initial splits takes more time than processing all the files(40 Gb, 4 .bam files) entirely on my local machine(BTW splits computing takes only 10% CPU, since it is single-threaded process). When the actual reading starts, I get OutOfMemoryError out of OverlapDetector(I start java with limited memory). Of course for small .bed sizes (100-1000 intervals) this fix could be OK, since it probably speeds up iterating the reads, but at the cost of memory taken by OverlapDetector structures.

akiezun:
I tested on a >100k bed file and it was fine (more than 10x improvement). I
tested on a spark cluster and give 6GB to each executor. I agree this may
not be the last word on the problem but it's much better than the current
solution of iterating over all reads

ddurnev:
How long was it splitting your .BAM files(size in GB?) with 100K bed? 100K is number of intervals, i. e. lines, not .bed file size in bytes, right? For me it was smth more than 20 min(local, not spark cluster). Maybe we should open another issue for the splitting problem?

src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
```diff
@@ -224,25 +232,23 @@ public static long getKey0(int refIdx, int alignmentStart0) {
...
+	private boolean overlaps(SAMRecord r) {
 		if (intervals == null ||
 				(r.getReadUnmappedFlag() && r.getAlignmentStart() == SAMRecord.NO_ALIGNMENT_START)) {
 			return true;
 		}
-		for (Locatable interval : intervals) {
-			if (r.getReadUnmappedFlag()) {
+		if (r.getReadUnmappedFlag()) { // special case for unmapped reads with coordinate set
+			for (Locatable interval : intervals) {
 				if (interval.getStart() <= r.getStart() && interval.getEnd() >= r.getStart()) {
```
cmnbroad: One meta comment - I think the intervals we're getting from the user are intended to be [query intervals](https://github.com/samtools/htsjdk/blob/c847e260a351dd4ded3caf3c44c6ec7d0c037675/src/java/htsjdk/samtools/QueryInterval.java#L12), which means the interval end value can use 0 or -1 as a sentinel to mean the end of the reference sequence.  So this code path, for example, should return true for an unmapped, placed read located at [20, x] against the query interval [10, -1]. It doesn't look to me like the overlap detector is intended to handle these either. We should probably add tests for these cases.

The code in htsjdk for handling these is a bit distributed, but there is code in the QueryInterval class as well as [here](https://github.com/samtools/htsjdk/blob/dfe2b7873beec5ce1bc9ebb410dbc6f05f6bbcda/src/java/htsjdk/samtools/BAMFileReader.java#L944) that illustrates how BAM index queries handle this.

src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
```diff
@@ -164,6 +167,11 @@ public static long getKey0(int refIdx, int alignmentStart0) {
 		readPair = false;
 		lastOfPair = false;
 		intervals = BAMInputFormat.getIntervals(conf);
+		if (intervals != null) {
+			overlapDetector = new OverlapDetector<>(0, 0);
+			overlapDetector.addAll(intervals, intervals);
```
cmnbroad: It might be worth [optimizing](https://github.com/samtools/htsjdk/blob/c847e260a351dd4ded3caf3c44c6ec7d0c037675/src/java/htsjdk/samtools/QueryInterval.java#L73) these intervals before handing them to the overlap detector.

cmnbroad:
@tomwhite Back to you - I had a couple of comments/questions about whether we're handling interval queries correctly - I think we need to for these cases ?

AndreSchumacher:
I was curious about the performance impact and did a little micro-benchmark. Below a screenshot from VisualVM's CPU time sampler. This is for a smallish BAM (2GB) and an artificial BED file (10k intervals).
visual_vm_read_bam_filtered_by_interval

In this case just reading the full file is a lot faster because the overhead of the filtering is larger than the benefit because the BED file intervals are a lot finer than the BAM file split points. One solution could be to come up with a heuristic that estimates when filtering is better than just reading the complete file or just clearly state that this functionality is for a smallish number of query intervals only. What do you guys think?

ddurnev:
@AndreSchumacher - this is exactly the point I'm talking about. But I want to stress that filtering splits by interval takes so much time in case of number of intervals is large(10s of thousands) even when it is useful, for example for big .BAM files, say 250 Gb. Could quering of the index for every single interval be optimized? Maybe we could somehow merge close small intervals into some bigger parts before quering?

tomwhite:
I agree that this can be improved for large numbers of intervals. The intervals are used in two places: i) to trim or eliminate input splits, and ii) to filter out reads.

@ddurnev and @AndreSchumacher, I think you are mainly concerned with i), right? Certainly when there are a large number of intervals that are spread throughout the input then i) is not really needed and it may be faster to use the original splits unchanged. One option would be to ignore the intervals for trimming/eliminating splits if there are more than a certain number (1000?), and emit a log message. This wouldn't affect correctness of course.

For ii), we could do better by - for each split - throwing away any interval that doesn't overlap the split. This would save on memory (in the reader) and avoid keeping any intervals around that are never going to match.

Would it be OK to address these improvements in future PRs?

@cmnbroad thanks for the review! I will post an updated PR next week (travelling at the moment).

AndreSchumacher:
@tomwhite sure, let's address these in later PR's. Agreed, a simple rule to give up and warn on encountering a larger number of intervals may be the easiest and most robust solution and yes I was thinking of i). In my test it seems that most time was actually spent computing the chunks for the set of intervals. The PR looks otherwise good to me.

tomwhite:
@cmnbroad I'm not sure we do support 0/-1 semantics to mean end of reference sequence - as far as I can tell SAMtools `-L` (for BED files) doesn't, which is what we're equivalent to.

We could probably add it though, so I had a look to see what it would involve. `QueryInterval` doesn't implement `htsjdk.samtools.util.Locatable`, so we'd have to have a separate `setQueryIntervals` method on `BAMInputFormat`. We'd then convert a `QueryInterval` into an `Interval` by changing 0/-1 to end of sequence (using the BAM header to get the sequence dictionary), and then we'd be able to use `OverlapDectector`. This seems like an enhancement, so I'd prefer to do it in a separate PR.

cmnbroad:
Resolving the QueryInterval issue in another PR may be best. I see now that the Hadoop-BAM doc advertises that it supports samtools -L style, which I interpret to be BED 0-based intervals, but I think GATK (and the htsjdk SamReader index query APIs) mostly use 1-based "Picard" style intervals. And I see conflicting evidence about whether GATK needs the QueryInterval style matching (which is what the htsjdk SamReader index query APIs use).

tomwhite:
Thanks @cmnbroad. If you agree @akiezun, then I think we can merge this.

akiezun:
I agree that other enhancements should be in new PRs. To clarify, because i got lost in the comments a bit, does this PR always improve performance? If yes, let's just merge it. If not, what should be done then (is there a heuristic to detect those cases?)

tomwhite:
This code is an improvement. I've merged it and opened #91 and #92 for further optimizations and improvements.

PR #83, opened 2016-04-18
Removing old command-line interface

AndreSchumacher:
I confirmed with the other developers and it seems the command-line interface is no longer needed and in fact very much outdated, so let's remove it.

PR #84, opened 2016-04-20
Remove CmdLineParser since the CLI has been removed now (in #83).

tomwhite:


AndreSchumacher:
@tomwhite: I was about to revise my PR but now you did this already which is great :-) go ahead, let's clean this up. Thanks, I completely had forgotten about this "inline" dependency.

PR #86, opened 2016-04-21
Remove support for Hadoop 1

### LINKED ISSUE #54: Remove support for Hadoop 1
There hasn't been a release of Hadoop 1 since 2013, and all the distributions have been based on Hadoop 2 for a long time now.

This would involve removing references to Hadoop 1 in the documentation, as well as removing the `hadoop.org.apache.hadoop.mapreduce.lib` packages (which seem to be unused?). It would be good to mark the Hadoop dependencies as `provided`, although this could be done separately.

tomwhite:
* Remove the hadoop1 Maven profile.
* Remove ContextUtil and call Hadoop Configuration/Context API directly.

Fixes #54.


pom.xml
```diff
@@ -50,8 +50,6 @@
 
         <hadoop.version>2.0.0-cdh4.6.0</hadoop.version>
         <hadoop.version>2.0.0-alpha</hadoop.version>
-        <hadoop.version>1.2.1</hadoop.version>
```
AndreSchumacher: For completeness, could we remove the same lines from examples/pom.xml?

AndreSchumacher:
This looks good to me. Excellent that we can remove the ContextUtil now.

tomwhite:
Thanks @AndreSchumacher. I removed the corresponding lines from the example pom. Merging now.

PR #87, opened 2016-04-21
Deprecate unused utility classes

tomwhite:
This deprecates Pair, Timer, and SeekableArrayStream since they are no longer used internally.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/util/Pair.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/util/Pair.java
@@ -22,6 +22,10 @@
 
 package org.seqdoop.hadoop_bam.util;
 
+/**
+ * @deprecated
+ */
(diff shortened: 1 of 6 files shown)


src/main/java/org/seqdoop/hadoop_bam/util/Pair.java
```diff
@@ -22,6 +22,10 @@
 
 package org.seqdoop.hadoop_bam.util;
 
+/**
+ * @deprecated
+ */
```
heuermh: Could you add a note about when these will be removed?  Something like
```This class will be removed in the next major version release (8.0).```

tomwhite:
Thanks for looking @heuermh. I've added the text you suggested.

heuermh:
LGTM, thanks @tomwhite

PR #90, opened 2016-04-27
Improve block/gzip compressed VCF format detection

tomwhite:


heuermh:
This looks good!  I was working on something more complicated with BlockCompressedInputStream but duh BGZF files are already gzip.

tomwhite:
@heuermh does this fix the problem you were hitting?

heuermh:
Yes, the only unit tests failing here for us are due to lack of BCF support in htsjdk
https://github.com/bigdatagenomics/adam/pull/989

Thank you!

PR #94, opened 2016-05-13
Upgrade to htsjdk 2.3.0

tomwhite:
Fix test due to changes in htsjdk (can't index a queryname sorted bam)

Use synchronous streams for SamReaderFactory to avoid readahead problems when underlying stream is accessed directly.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java
@@ -108,10 +108,13 @@ public class BAMSplitGuesser extends BaseSplitGuesser {
 		throws IOException
 	{
 		// Use a reader to skip through the headers at the beginning of a BAM file, since
-		// the headers may exceed MAX_BYTES_READ in length
+		// the headers may exceed MAX_BYTES_READ in length. Don't close the reader
+		// otherwise it will close the underlying stream, which we continue to read from
+		// on subsequent calls to this method.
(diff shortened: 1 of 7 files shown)


src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java
```diff
@@ -111,7 +111,8 @@ public long guessNextBAMRecordStart(long beg, long end)
 		// the headers may exceed MAX_BYTES_READ in length
 		if (beg == 0) {
 			this.inFile.seek(beg);
-			SamReader open = SamReaderFactory.makeDefault().open(SamInputResource.of(inFile));
+			SamReader open = SamReaderFactory.makeDefault().setUseAsyncIo(false)
```
akiezun: @tomwhite I think it would be good to have a comment here that this method does not close the stream, which means it can be used by subsequent calls to the same method. This will help future readers of this code

tomwhite: Good point. Fixed.

tomwhite:
Unless there are any objections, I'd like to merge this soon.

droazen:
@akiezun please review

heuermh:
LGTM

akiezun:
are we good for a release? I could use a new version of hadoop bam in gatk

tomwhite:
I think so. I'll try tomorrow.

PR #98, opened 2016-06-07
Test BCF inference from file path and data.

tomwhite:


tomwhite:
This is adding extra tests - merging.

PR #99, opened 2016-06-07
Throw PathNotFoundException if path not found in createRecordReader

### LINKED ISSUE #79: Improve error message in createRecordReader when file missing
`AnySAMInputFormat.createRecordReader` throws `IllegalArgumentException` if it can't determine the format of a sam/bam file. If this is caused by a nonexistant file it hides the FileNotFoundException which would be a more useful error message.

This cropped up in a case where a file local to one machine was being specified as an input a spark cluster with multiple nodes. The file existed for the master node, but the workers failed because it was unavailable to them.

It would be helpful if the error message mentioned this possibility.

see https://github.com/broadinstitute/gatk/issues/1417 for the original issue

tomwhite:
Fixes #79


tomwhite:
@akiezun, does this look OK to you?

PR #97, opened 2016-06-07
Fix VCFHeaderReader for gzipped and BGZF VCF formats

tomwhite:
Fixes #96

PR #106, opened 2016-06-15
Add BGZFCodec for reading and writing files with a .bgz suffix.

tomwhite:
This adds support for writing BGZF-compressed files.

The old BGZF codec has been renamed BGZFEnhancedGzipCodec and reads .gz files
as BGZF (so they are splittable) if possible, otherwise falls back to regular
gzip.


heuermh:
I'm generally in favor of this, though it may take us a while to test.

tomwhite:
I'd like this to go into the next release, so please take a look/try it out @heuermh.

src/test/java/org/seqdoop/hadoop_bam/TestVCFRoundTrip.java
```diff
@@ -0,0 +1,212 @@
...
+        final boolean success = job.waitForCompletion(true);
+        assertTrue(success);
+
+        return outputPath;
+    }
+
+    private static VCFFileReader parseVcf(File vcf) throws IOException {
+        File actualVcf;
+        // work around TribbleIndexedFeatureReader not reading header from .bgz files
+        if (vcf.getName().endsWith(".bgz")) {
```
heuermh: Is this only a workaround for unit tests, or something that we'll also need to do downstream?

heuermh:
+1, see downstream pull request referenced above

src/test/java/org/seqdoop/hadoop_bam/TestVCFRoundTrip.java
tomwhite: This is only needed for testing. VCFHeaderReader in Hadoop-BAM will correctly read a bgzf stream now (since #97).

heuermh: Great, thanks!  Yep, VCFHeaderReader is now working for us.

cmnbroad: @tomwhite - is that (.bgz/TribbleIndexedFeatureReader) a known issue or is there a ticket? It seems like an easy fix.

src/main/java/org/seqdoop/hadoop_bam/KeyIgnoringVCFOutputFormat.java
```diff
@@ -90,7 +95,23 @@ public void readHeaderFrom(SeekableStream in) throws IOException {
...
 		throws IOException
 	{
-		return getRecordWriter(ctx, getDefaultWorkFile(ctx, ""));
+		Configuration conf = ctx.getConfiguration();
+		boolean isCompressed = getCompressOutput(ctx);
+		CompressionCodec codec = null;
+		String extension = "";
+		if (isCompressed) {
+			Class<? extends CompressionCodec> codecClass =
+					getOutputCompressorClass(ctx, GzipCodec.class);
```
cmnbroad: GzipCodec seems like an unnatural default, certainly for BCF. Wouldn't BGZFCodec be a better choice for both for both VCF and BCF?

cmnbroad:
A couple of minor comments. Not sure why the Travis push job failed. Otherwise LGTM. Back to @tomwhite.

src/test/java/org/seqdoop/hadoop_bam/TestVCFRoundTrip.java
tomwhite: I just created https://github.com/samtools/htsjdk/issues/653 for it.

src/main/java/org/seqdoop/hadoop_bam/KeyIgnoringVCFOutputFormat.java
tomwhite: Agreed. Fixed.

tomwhite:
Thanks for taking a look @heuermh and @cmnbroad. I think the earlier Travis job failed because I pushed the change to the wrong upstream repository then deleted it before the job had a chance to complete. It's passing now. I'll merge this shortly.

heuermh:
Thank you, @tomwhite!

Issue #61, opened 2016-02-03
SAMOutputPreparer should be extended to write complete streams


cmnbroad:
SAMOutputPreparer knows how to prepare the beginning of an output stream, but not how to populate or terminate it (which for CRAM requires a special terminating EOF container). We should extend it do the whole roll-up correctly based on the SAMFormat being used.

tomwhite:
Fixed in #111.

Issue #60, opened 2016-02-03
Make Utils#mergeInto more configurable


tomwhite:
This is so it can be used directly from GATK (which currently replicates its functionality).

cmnbroad:
This also needs to be updated to handle CRAM files.

tomwhite:
Addressed in #111.

Issue #48, opened 2016-01-05
Generate splitting-bai indexes for BAM when writing from MR


tomwhite:
This would avoid the need to use heuristics to find splits when reading back in.

tomwhite:
This should also handle merging the indexes into one when the BAM shards are merged. See https://github.com/HadoopGenomics/Hadoop-BAM/issues/60

tomwhite:
Addressed in https://github.com/HadoopGenomics/Hadoop-BAM/commit/cd89366d86302ba7d0c02079fc3d663dd46408ce

PR #111, opened 2016-07-06
Write splitting BAM indexes, and code for merging sharded BAM or CRAM files

tomwhite:
This moves code from GATK to Hadoop-BAM for merging sharded BAM or CRAM files. It also adds an option to write splitting BAM indexes from MR (defaults to off). The splitting indexes (if present) are merged into a single index as a part of the merge operation.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndexer.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndexer.java
@@ -194,6 +194,13 @@ public final class SplittingBAMIndexer {
 		count++;
 	}
 
+	void processAlignment(final long virtualOffset) throws IOException {
+		if (count == 0 || (count + 1) % granularity == 0) {
+			writeVirtualOffset(virtualOffset);
+		}
+		count++;
+	}
+
 	private long getPos(SAMFileSpan filePointer) {
 		// Use reflection since BAMFileSpan is package private in htsjdk 1.141. Note that
 		// Hadoop-BAM cannot use a later version of htsjdk since it requires Java 8.
--- a/src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndexer.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndexer.java
@@ -221,7 +221,12 @@ public final class SplittingBAMIndexer {
 		}
 	}
 
-	private void writeVirtualOffset(long virtualOffset) throws IOException {
+	/**
+	 * Write the given virtual offset to the index. This method is for internal use only.
+	 * @param virtualOffset virtual file pointer
+	 * @throws IOException
+	 */
+	public void writeVirtualOffset(long virtualOffset) throws IOException {
 		lb.put(0, virtualOffset);
 		out.write(byteBuffer.array());
 	}
--- a/src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
@@ -18,6 +18,8 @@ class NIOFileUtil {
   private NIOFileUtil() {
   }
 
+  public static final String PARTS_GLOB = "glob:**/part-[mr]-[0-9][0-9][0-9][0-9][0-9]*";
+
   /**
    * Convert the given path string to a {@link Path} object.
    * @param path the path to convert
--- a/src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
@@ -1,0 +1,88 @@
+package org.seqdoop.hadoop_bam.util;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.URI;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+class NIOFileUtil {
+  private NIOFileUtil() {
+  }
+
+  /**
+   * Convert the given path string to a {@link Path} object.
+   * @param path the path to convert
+   * @return a {@link Path} object
+   */
+  public static Path asPath(String path) {
+    URI uri = URI.create(path);
+    return uri.getScheme() == null ? Paths.get(path) : Paths.get(uri);
+  }
+
+  /**
+   * Delete the given directory and all of its contents if non-empty.
+   * @param directory the directory to delete
+   * @throws IOException
+   */
+  public static void deleteRecursive(Path directory) throws IOException {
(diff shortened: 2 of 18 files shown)


tomwhite:
Addresses #48, #60, and #61.

fnothaft:
+1, we have similar code in ADAM, will be good to eliminate it.

tomwhite:
@fnothaft would you (or someone else on the ADAM team) be able to review this? I tried running GATK with this change and it passes tests - would be good for someone to do the same with ADAM.

fnothaft:
@tomwhite I'm traveling this week so I may be a bit slow, but I should be able to make a pass by the end of the week. Would that timing be OK for you?

tomwhite:
@fnothaft, that would be great - thanks!

tomwhite:
I've updated this so that it uses the Java NIO file API for metadata and merging operations (this allows it to be used from GATK which supports other NIO implementations). I've also added merge for VCF files.

@fnothaft can you take a look please? @cmnbroad, it would be great to get your feedback too.

src/main/java/org/seqdoop/hadoop_bam/util/SAMFileMerger.java
```diff
@@ -0,0 +1,150 @@
...
+
+    List<Path> parts = getFilesMatching(partPath,
+        "glob:**/part-[mr]-[0-9][0-9][0-9][0-9][0-9]*",
+        SplittingBAMIndexer.OUTPUT_FILE_EXTENSION);
+    if (parts.isEmpty()) {
+      throw new IllegalArgumentException("Could not write bam file because no part " +
+          "files were found in " + partPath);
+    }
+
+    Files.deleteIfExists(outputPath);
```
cmnbroad: Should we be deleting the index here as well ?

src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
```diff
@@ -0,0 +1,88 @@
...
+   * @param out the stream to write each file into, in order
+   * @throws IOException
+   */
+  public static void mergeInto(List<Path> parts, OutputStream out)
+      throws IOException {
+    for (final Path part : parts) {
+      Files.copy(part, out);
+    }
+    for (final Path part : parts) {
+      Files.delete(part);
```
cmnbroad: It looks like (both) mergeParts implementations delete these at the end as well, so I'm wondering if we need this here. Doing it at the end after everything succeeds makes sense to me.

src/main/java/org/seqdoop/hadoop_bam/util/SAMFileMerger.java
```diff
@@ -0,0 +1,150 @@
...
+      writeTerminatorBlock(out, samOutputFormat);
+    }
+    long fileLength = Files.size(outputPath);
+
+    final Path outputSplittingBaiPath = outputPath.resolveSibling(
+        outputPath.getFileName() + SplittingBAMIndexer.OUTPUT_FILE_EXTENSION);
+    try (final OutputStream out = Files.newOutputStream(outputSplittingBaiPath)) {
+      mergeSplittingBaiFiles(out, partPath, headerLength, fileLength);
+    } catch (IOException e) {
+      deleteRecursive(outputSplittingBaiPath);
```
cmnbroad: I think this should rethrow, otherwise we're masking this failure.

src/main/java/org/seqdoop/hadoop_bam/util/SAMFileMerger.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+    final Path partPath = asPath(partDirectory);
+    final Path outputPath = asPath(outputFile);
+    if (partPath.equals(outputPath)) {
+      throw new IllegalArgumentException("Cannot merge parts into output with same " +
+          "path: " + partPath);
+    }
+
+    List<Path> parts = getFilesMatching(partPath,
+        "glob:**/part-[mr]-[0-9][0-9][0-9][0-9][0-9]*",
```
cmnbroad: This glob string should be a constant somewhere as its repeated in several places.

src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndexer.java
```diff
@@ -194,6 +194,13 @@ public void processAlignment(final SAMRecord rec) throws IOException {
 		count++;
 	}
 
+	void processAlignment(final long virtualOffset) throws IOException {
```
cmnbroad: This doesn't appear to be used anywhere?

src/main/java/org/seqdoop/hadoop_bam/util/VCFFileMerger.java
```diff
@@ -0,0 +1,127 @@
...
+public class VCFFileMerger {
+  /**
+   * Merge part file shards produced by {@link KeyIgnoringVCFOutputFormat} into a
+   * single file with the given header.
+   * @param partDirectory the directory containing part files
+   * @param outputFile the file to write the merged file to
+   * @param header the header for the merged file
+   * @throws IOException
+   */
+  public static void mergeParts(final String partDirectory, final String outputFile,
```
cmnbroad: This fails with .bcf files (even a legitimate .bcf that htsjdk can handle, and that passes the testRoundTrip test, fails the testRoundTripWithMerge test). I'm guessing its not a quick fix though, so we might want to make a separate ticket, and for now just reject them here and document the restriction.

src/main/java/org/seqdoop/hadoop_bam/util/SAMFileMerger.java
tomwhite: Yes. Fixed.

src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
tomwhite: This method deletes the part files after they have been merged, so this makes sense here. The directory that contained the part files (and also metadata like _SUCCESS, and index files) is deleted after the whole operation succeeded.

src/main/java/org/seqdoop/hadoop_bam/util/SAMFileMerger.java
tomwhite: Done.

src/main/java/org/seqdoop/hadoop_bam/util/SAMFileMerger.java
tomwhite: Done

src/main/java/org/seqdoop/hadoop_bam/SplittingBAMIndexer.java
tomwhite: It's used in BAMRecordWriter to write splitting BAM indexes from MR. See https://github.com/HadoopGenomics/Hadoop-BAM/pull/111/commits/cd89366d86302ba7d0c02079fc3d663dd46408ce#diff-8ce0e02bf18a51ccfbbe62aebc2f9498R141.

src/main/java/org/seqdoop/hadoop_bam/util/VCFFileMerger.java
tomwhite: I've added a check so that BCF files are rejected.

tomwhite:
Thanks for the review @cmnbroad! I've addressed your feedback in a new commit.

cmnbroad:
LGTM.

PR #114, opened 2016-08-26
honor AnySAMInputFormat.TRUST_EXTS_PROPERTY

lbergelson:
AnySAMInputFormat was setting trustExtensions only when it was constructed with a configuration
this was causing the property to be ignored when the configuration was being passed in later like what's done in GATK

removed the trustExtensions field and replaced it with a check of the configuration property at the time of use
  now TRUST_EXTS_PROPERTY will be honored no matter when the configuration is set

added a new private constructor and made all existing constructors delegate to that one in order to decrease confusion.

This means that by default the file extensions will be trusted like the documentation said.  The fact that it wasn't (and couldn't be enabled) was making it impossible to read headerless SAM files in GATK.


lbergelson:
@tomwhite Could you review this when you get a chance?  It's needed to fix https://github.com/broadinstitute/gatk/issues/2119

src/main/java/org/seqdoop/hadoop_bam/AnySAMInputFormat.java
```diff
@@ -40,6 +34,12 @@
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
 import org.apache.hadoop.mapreduce.lib.input.FileSplit;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
```
tomwhite: Avoid unecessary import reordering.

src/main/java/org/seqdoop/hadoop_bam/AnySAMInputFormat.java
```diff
@@ -68,9 +68,8 @@
 	private final boolean              givenMap;
 
 	private Configuration conf;
-	private boolean trustExts;
 
-	/** Creates a new input format, which will use the
+    /** Creates a new input format, which will use the
```
tomwhite: Whitespace

src/test/java/org/seqdoop/hadoop_bam/TestAnySAMInputFormat.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class TestAnySAMInputFormat {
+
+    @Test
+    public void testHeaderlessSamFormat() throws PathNotFoundException {
+        final SAMFormat result = getSamFormat(new Configuration(), "test_headerless.sam");
+        assertEquals(result, SAMFormat.SAM);
```
tomwhite: Flip the arguments here, since for JUnit (unlike TestNG) the first argument should be the expected value, and the second the actual one.

src/test/java/org/seqdoop/hadoop_bam/TestAnySAMInputFormat.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Test
+    public void testTrustExtensionsIsHonored() throws PathNotFoundException {
+        final Configuration conf = new Configuration();
+        //default to trusting exceptions
+        assertEquals(getSamFormat(conf, "misnamedBam.sam"), SAMFormat.SAM);
```
tomwhite: Flip

src/test/java/org/seqdoop/hadoop_bam/TestAnySAMInputFormat.java
```diff
@@ -0,0 +1,33 @@
...
+
+        conf.set(AnySAMInputFormat.TRUST_EXTS_PROPERTY, "false");
+        final SAMFormat result = getSamFormat(conf, "misnamedBam.sam");
+        assertEquals(result, SAMFormat.BAM);
```
tomwhite: Flip

src/main/java/org/seqdoop/hadoop_bam/AnySAMInputFormat.java
lbergelson: ack, auto correct struck again.

tomwhite:
Thanks for the contribution @lbergelson. Looks fine apart from a few minor comments.

Could you also update the documentation as it currently says that the default for `hadoopbam.anysam.trust-exts` is false (see https://github.com/HadoopGenomics/Hadoop-BAM#library-usage, second table).

src/test/java/org/seqdoop/hadoop_bam/TestAnySAMInputFormat.java
lbergelson: Thanks for pointing that out.  I didn't realize.  What a strange decision for TestNG to have made to change the api like that.

lbergelson:
@tomwhite I've addressed the comments and squashed.

PR #120, opened 2016-10-20
Handle FileSystemNotFoundException

tomwhite:
This is needed when the filesystem provider is loaded using a URL classloader (e.g. in spark-submit).

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/CRAMRecordWriter.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/CRAMRecordWriter.java
@@ -15,6 +15,7 @@ import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.mapreduce.RecordWriter;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 
+import org.seqdoop.hadoop_bam.util.NIOFileUtil;
 import org.seqdoop.hadoop_bam.util.SAMHeaderReader;
 
 /** A base {@link RecordWriter} for CRAM records.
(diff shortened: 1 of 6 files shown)


src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
```diff
@@ -11,14 +13,36 @@
...
+  public static Path asPath(URI uri) {
+    try {
+      return Paths.get(uri);
+    } catch (FileSystemNotFoundException e) {
+      ClassLoader cl = Thread.currentThread().getContextClassLoader();
+      if (cl == null) {
+        throw e;
+      }
+      try {
+        return FileSystems.newFileSystem(uri, new HashMap<>(), cl).provider().getPath(uri);
```
cmnbroad: Is there any way to write a for this code path ? Its not obvious to me since it seems like it would require a reference to a FS that isn't instantiated ?

src/main/java/org/seqdoop/hadoop_bam/CRAMRecordWriter.java
```diff
@@ -78,10 +79,9 @@ private void init(
         origOutput = output;
         this.writeHeader = writeHeader;
 
-        final URI referenceURI = URI.create(
-                ctx.getConfiguration().get(CRAMInputFormat.REFERENCE_SOURCE_PATH_PROPERTY)
-        );
-        refSource = new ReferenceSource(Paths.get(referenceURI));
+        final String referenceURI =
+                ctx.getConfiguration().get(CRAMInputFormat.REFERENCE_SOURCE_PATH_PROPERTY);
+        refSource = new ReferenceSource(NIOFileUtil.asPath(referenceURI));
```
tomwhite: Fixed.

src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
tomwhite: This is tricky to write a unit test for. I've tested it manually though, by using spark-submit (where it failed without the change), so that should be sufficient.

tomwhite:
Thanks for the review @cmnbroad! I'd like to release this in the next day or so.

Issue #91, opened 2016-05-05
Improve performance for large interval lists


tomwhite:
A couple of ideas from #82:
1.  Ignore the intervals for trimming/eliminating splits if there are more than a certain number (1000?), and emit a log message. (The intervals would still be used for filtering of course.)
2. For each split, discard away any interval that doesn't overlap the split. This would be done by the record reader.

tomwhite:
I tried #2, but found that it actually makes things slower. Presumably because checking if each interval overlaps the split is slow, compared to the optimized checking that OverlapDetector performs.

tomwhite:
Fixed in #121.

lbergelson:
Is the performance with intervals still worse than the performance without interval lists?

droazen:
If yes, I think we might want to keep this issue open

tomwhite:
@lbergelson, yes it is.

As it currently stands running CountReadsSpark (from GATK) on a 133 GB BAM takes 2.17 minutes when no intervals are specified, and 7.26 minutes when a human exome interval list (180K intervals) is used. This is over 3 times slower. It's a lot faster than it was (due to the fix in #121), but we can do better.

The reason that it is slower when using a long intervals list is because the filtering happens after each read has been read from the BAM. If we could do the filtering by skipping portions of the BAM file that don't overlap (like HTSJDK does) then we can do a lot better.

I mocked up some code to do this, and the time was 1.73 minutes - 20% faster than the no intervals case.

See https://github.com/HadoopGenomics/Hadoop-BAM/compare/master...tomwhite:intervals-optimization-experiment

* The main idea is to add the chunk intervals for each split into the FileVirtualSplit class, so that each record reader can use that information to only read the chunks in the BAM file that it needs to, rather than read everything and apply filtering afterwards.
* This change means that an index is always needed when using intervals. I think that's probably reasonable.
* BAMRecordReader changes to use an iterator from SamReader, rather than manage its own iteration (via BAMRecordCodec).
* There is no provision in htsjdk to access the underlying reader, in this case BAMFileReader, and in particular seek to a known point in the block compressed stream (BGZF stream). This is needed by Hadoop-BAM to process splits in parallel. I had to expose some parts of BAMFileReader to get this to work. See https://github.com/samtools/htsjdk/compare/master...tomwhite:tw_expose_underlying_sam_reader
[...]

PR #121, opened 2016-12-14
Improve performance for large interval lists

tomwhite:
Ignore the intervals for trimming/eliminating splits if there are more than a given number, specified by the property `hadoopbam.bam.max-intervals-for-split-filtering`, default 1000.

This addresses #91

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
@@ -25,13 +25,12 @@ package org.seqdoop.hadoop_bam;
 import htsjdk.samtools.BAMFileSpan;
 import htsjdk.samtools.BAMIndex;
 import htsjdk.samtools.Chunk;
-import htsjdk.samtools.DiskBasedBAMFileIndex;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMSequenceDictionary;
-import htsjdk.samtools.SamFiles;
+import htsjdk.samtools.SamReader;
+import htsjdk.samtools.SamReaderFactory;
 import htsjdk.samtools.util.Interval;
 import htsjdk.samtools.util.Locatable;
-import java.nio.file.Paths;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -289,8 +298,14 @@ private int addProbabilisticSplits(
 	private List<InputSplit> filterByInterval(List<InputSplit> splits, Configuration conf)
 			throws IOException {
 		List<Interval> intervals = getIntervals(conf);
+		int maxIntervalsForSplitFiltering = conf.getInt(MAX_INTERVALS_FOR_SPLIT_FILTERING_PROPERTY,
+				DEFAULT_MAX_INTERVALS_FOR_SPLIT_FILTERING);
 		if (intervals == null) {
 			return splits;
+		} else if (intervals.size() > maxIntervalsForSplitFiltering) {
+			System.err.println("INFO: not filtering splits by intervals since number of " +
```
lbergelson: Does hadoop bam have a logging framework?  Better to emit a logging message that could be configured by downstream projects than just the system.out dump.

droazen:
So, to clarify what this does: if there are more than 1000 intervals, you'll give up and load the entire bam, then do filtering after the fact?

lbergelson:
@tomwhite I'm a bit worried that this approach is too simple. There are definitely use cases that generate many intervals but only collectively cover a small portion of the genome.  One common one would be look at all chromosome 20 variant sites and their overlapping reads.  In that case we'd probably still want to do large scale filtering before reading the bam, since all the intervals are limited to a small subset of the bam, but we'd still have thousands of intervals.  We should definitely merge this now to help with the exome interval list case,  but we might want to consider a more nuanced approach for the future.

lbergelson:
Is the slow down because of the splits * chunks comparison?  We could use a more efficient approach to do that calculation if we needed to.

tomwhite:
@lbergelson yes, that's the heart of the problem - the large number of chunks. For a file with a large number of intervals that I tried (180 thousand), it produced around 2 million chunks. After I tried optimizing the list of chunks, it came down to 43 thousand, and the split computation took seconds. The other change I made was to use a BAM reader that caches the index file, so finding the initial set of chunks is fast (seconds, rather than minutes).

With these changes I don't think we need the extra property that I originally added.

lbergelson:
@tomwhite This seems like a much better solution!  The previous way seemed kind of hacky and like it would have weird edge cases.  👍  from me.

PR #122, opened 2017-01-19
Use htsjdk machinery to make reading with intervals significantly faster

tomwhite:
The main idea is to add the chunk intervals for each split into
the FileVirtualSplit class, so that each record reader can use
that information to only read the chunks in the BAM file that it
needs to, rather than read everything and apply filtering afterwards.

This change means that an index is always needed when using
intervals.

BAMRecordReader changes to use an iterator from SamReader, rather
than manage its own iteration (via BAMRecordCodec).

Support for keeping paired reads together in the same split has
been removed (previously set via
'hadoopbam.bam.keep-paired-reads-together'), since it is not
compatible with BAMFileReader managing the chunks to iterate over.
Applications that require this functionality can use the technique
in GATK's BwaSparkEngine#putPairsInSamePartition to achieve the
same result.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
@@ -22,9 +22,11 @@
 
 package org.seqdoop.hadoop_bam;
 
+import htsjdk.samtools.BAMFileReader;
 import htsjdk.samtools.BAMFileSpan;
 import htsjdk.samtools.BAMIndex;
 import htsjdk.samtools.Chunk;
+import htsjdk.samtools.QueryInterval;
 import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMSequenceDictionary;
 import htsjdk.samtools.SamReader;
@@ -32,7 +34,9 @@ import htsjdk.samtools.SamReaderFactory;
 import htsjdk.samtools.util.Interval;
 import htsjdk.samtools.util.Locatable;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.Map;
 import java.util.Set;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileSystem;
@@ -67,16 +71,6 @@ public class BAMInputFormat
 {
 	// set this to true for debug output
 	public final static boolean DEBUG_BAM_SPLITTER = false;
-
-	/**
-	 * If set to <code>true</code>, ensure that for paired reads both reads in a pair are
-	 * always in the same split for queryname-sorted BAM files.
-	 * <p>
-	 * Note: only use this option if for all paired reads both reads in each pair are
-	 * present, otherwise it is possible that reads may be silently dropped.
-	 */
-	public static final String KEEP_PAIRED_READS_TOGETHER_PROPERTY =
-			"hadoopbam.bam.keep-paired-reads-together";
 
 	/**
 	 * Filter by region, like <code>-L</code> in SAMtools. Takes a comma-separated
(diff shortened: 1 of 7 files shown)


fnothaft:
> Support for keeping paired reads together in the same split has been removed (previously set via 'hadoopbam.bam.keep-paired-reads-together')

We're starting to make use of this functionality downstream in ADAM, but we have no plans to make use of the `BwaSparkEngine`. Is there a way that we can not lose this functionality? Performance when reading with a large number of intervals is not (currently) an important use case for us.

tomwhite:
@fnothaft - thanks for the heads up. I think this feature actually belongs at the application level, rather than in Hadoop-BAM, since it makes things more complicated in the lower level due to the way it interacts with things like intervals.

Also, these changes in Hadoop-BAM are not only beneficial for large numbers of intervals - this implementation is more performant even for a small number of intervals if the number of reads being filtered is significant, since the filtering is being done by only reading the intervals of interest from disk (rather than reading everything, then filtering).

I was inspired by @ryan-williams elegant https://github.com/hammerlab/magic-rdds for the tricks to move records from one partition to another in an arbitrary RDD. This is exactly what you need to ensure that paired reads are being kept together. The [code I wrote in BwaSparkEngine](https://github.com/broadinstitute/gatk/blob/tw_intervals_optimization/src/main/java/org/broadinstitute/hellbender/tools/spark/bwa/BwaSparkEngine.java#L89-L117) is essentially a more specialized version of Ryan's code for BorrowElemsRDD. I wasn't suggesting that you use BwaSparkEngine directly, but that you use the approach there, or BorrowElemsRDD. Would that work for you?

fnothaft:
Hi @tomwhite! Thanks for the quick and comprehensive reply! Let me take a more detailed look over your comments and the code that you linked to and get back to you. That's good to know about the changes being beneficial for a small number of intervals, because that is indeed an important use case for one of our projects. Thanks for a pointer to the specific `BwaSpark` method; that code does indeed seem fairly lightweight, so that would probably be a reasonable approach for us. In any case, let me take a look and I'll circle back later today or tomorrow.

droazen:
@cmnbroad Can you review?

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -308,60 +303,79 @@ private int addProbabilisticSplits(
 					.setUseAsyncIo(false);
 			SamReader samReader = readerFactory.open(NIOFileUtil.asPath(fs.makeQualified(bamFile).toUri()));
 			if (!samReader.hasIndex()) {
-				System.err.println("WARNING: no BAM index file found, splits will not be " +
-						"filtered, which may be very inefficient: " + bamFile);
-				return splits;
+				throw new IllegalArgumentException("Intervals set but no BAM index file found for " + bamFile);
```
cmnbroad: I would suggest to lift the reader factory construction out of the loop, and acquire the reader in a try-with resources.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -308,60 +303,79 @@ private int addProbabilisticSplits(
...
-
-			for (Locatable interval : intervals) {
-				int referenceIndex = dict.getSequenceIndex(interval.getContig());
-				int intervalStart = interval.getStart();
-				int intervalEnd = interval.getEnd();
-				BAMFileSpan sfs = idx.getSpanOverlapping(referenceIndex, intervalStart, intervalEnd);
-				chunks.addAll(sfs.getChunks());
-			}
+			QueryInterval[] queryIntervals = prepareQueryIntervals(intervals, dict);
+			fileToSpan.put(bamFile, BAMFileReader.getFileSpan(queryIntervals, idx));
```
cmnbroad: And also acquire the stream inside of a try with resources, since I think almost all of these methods can wind up throwing.

src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
```diff
@@ -126,62 +127,74 @@ public static long getKey0(int refIdx, int alignmentStart0) {
...
                 	System.err.println("XXX inizialized BAMRecordReader byte offset: " +
 				fileStart + " record offset: " + recordStart);
 		}
 
-		keepReadPairsTogether = SortOrder.queryname.equals(header.getSortOrder()) &&
-			conf.getBoolean(BAMInputFormat.KEEP_PAIRED_READS_TOGETHER_PROPERTY, false);
-		readPair = false;
-		lastOfPair = false;
-		intervals = BAMInputFormat.getIntervals(conf);
+		List<Interval> intervals = BAMInputFormat.getIntervals(conf);
```
cmnbroad: I'm wondering if we should throw if we see a call that has the KEEP_PAIRED_READS_TOGETHER_PROPERTY set, since otherwise we're silently no longer honoring it.

src/main/java/org/seqdoop/hadoop_bam/FileVirtualSplit.java
```diff
@@ -79,15 +88,32 @@ public FileVirtualSplit(Path f, long vs, long ve, String[] locs) {
 	public void setStartVirtualOffset(long vo) { vStart = vo; }
 	public void   setEndVirtualOffset(long vo) { vEnd   = vo; }
 
+	public long[] getIntervalFilePointers() {
```
cmnbroad: It would be helpful to have some comments in this file saying what these are (i.e., BAMFileSpan chunk start/stop pairs for the split), and that this will return null if no query interval was specified, etc.

src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
```diff
@@ -126,62 +127,74 @@ public static long getKey0(int refIdx, int alignmentStart0) {
...
 		final Path             file  = split.getPath();
 		final FileSystem       fs    = file.getFileSystem(conf);
 
-		this.stringency = SAMHeaderReader.getValidationStringency(conf);
+		ValidationStringency stringency = SAMHeaderReader.getValidationStringency(conf);
 
-		final FSDataInputStream in = fs.open(file);
+		java.nio.file.Path index = SamFiles.findIndex(NIOFileUtil.asPath(fs.makeQualified(file).toUri()));
+		Path fileIndex = index == null ? null : new Path(index.toUri());
+		SeekableStream inStream = fileIndex == null ? null : WrapSeekable.openPath(fs, fileIndex);
```
cmnbroad: If this were called indexStream it would be a lot clearer. Also this is another good try-with-resource candidate.

tomwhite:
Thanks for the review @cmnbroad. I've addressed your feedback.

PR #123, opened 2017-02-22
Improve creation of child path. 

tomwhite:
This change is needed to allow the merge logic to work on GCS.


droazen:
:+1: merge when ready @tomwhite

PR #124, opened 2017-03-19
fix FastaInputFormat infinite loop

ryan-williams:
Fastas that generate more than one split in `FileInputFormat` caused `FastaInputFormat` to infinite loop due to a `for` loop that didn't increment its loop variable.

This affected fasta files larger than 32MB, 64MB, or 128MB (depending on cluster defaults, and modulo the 10% "slop" margin).

[It's been this way since 2013, apparently](https://github.com/HadoopGenomics/Hadoop-BAM/commit/9b677c1f789ee46e14afedeb006de014397c9e6c#diff-20f7f979756780ec5e84f811ef1e8a20R86) ¯\_(ツ)_/¯.


tomwhite:
Good catch. +1

PR #130, opened 2017-05-05
Catch ProviderNotFoundException when picking BAM splits, and formally…

### LINKED ISSUE #125: Should codepaths that use java.nio catch ProviderNotFoundException?
I was running on a non-HDFS file system that did not have a registered `java.nio.file.spi.FileSystemProvider` for the FS scheme. I was loading the whole file, so I didn't _need_ the index, but was getting a `ProviderNotFoundException` when [loading the index when picking splits](https://github.com/HadoopGenomics/Hadoop-BAM/blob/master/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java#L172). Turns out, the BAM split picking code [only catches an IOError](https://github.com/HadoopGenomics/Hadoop-BAM/blob/master/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java#L152-L156) when using the BAM index. I think the correct behavior would be to also catch the `java.nio.file.ProviderNotFoundException` exception as well. Thoughts?

fnothaft:
… throw it from the RecordReader. Resolves #125.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
@@ -50,6 +50,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.nio.file.ProviderNotFoundException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -191,7 +192,7 @@ public class BAMInputFormat
 	/** Returns a {@link BAMRecordReader} initialized with the parameters. */
 	@Override public RecordReader<LongWritable,SAMRecordWritable>
 		createRecordReader(InputSplit split, TaskAttemptContext ctx)
-			throws InterruptedException, IOException
+            throws InterruptedException, IOException
 	{
 		final RecordReader<LongWritable,SAMRecordWritable> rr =
 			new BAMRecordReader();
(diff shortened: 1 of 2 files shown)


src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -114,7 +115,7 @@ static Path getIdxPath(Path path) {
 	/** Returns a {@link BAMRecordReader} initialized with the parameters. */
 	@Override public RecordReader<LongWritable,SAMRecordWritable>
 		createRecordReader(InputSplit split, TaskAttemptContext ctx)
-			throws InterruptedException, IOException
+            throws InterruptedException, IOException, ProviderNotFoundException, IllegalArgumentException
```
tomwhite: Both ProviderNotFoundException and IllegalArgumentException are runtime exceptions, so they shouldn't be added to the throws clause. Instead, they can be documented in javadoc to indicate the conditions under which they could be thrown by this method.

src/main/java/org/seqdoop/hadoop_bam/BAMRecordReader.java
```diff
@@ -117,7 +118,7 @@ public static long getKey0(int refIdx, int alignmentStart0) {
 	}
 
 	@Override public void initialize(InputSplit spl, TaskAttemptContext ctx)
-		throws IOException
+            throws IOException, ProviderNotFoundException, IllegalArgumentException
```
tomwhite: Same here.

tomwhite:
@fnothaft let me know if you'd like me to address the issues I mentioned.

fnothaft:
Hi @tomwhite! I've rebased this to clear the conflicts and addressed your comments.

PR #136, opened 2017-07-27
Add a way to get unmapped reads from a BAM.

### LINKED ISSUE #133: Add a way to get unmapped reads from a BAM
It's possible to filter by interval, but there's currently no way to include unmapped reads in the filter.

Needed for https://github.com/broadinstitute/gatk/issues/2571.

tomwhite:
Fixes #133.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
@@ -22,6 +22,8 @@
 
 package org.seqdoop.hadoop_bam;
 
+import com.google.common.collect.ImmutableList;
+import htsjdk.samtools.AbstractBAMFileIndex;
 import htsjdk.samtools.BAMFileReader;
 import htsjdk.samtools.BAMFileSpan;
 import htsjdk.samtools.BAMIndex;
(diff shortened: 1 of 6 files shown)


src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -72,31 +74,61 @@
 	// set this to true for debug output
 	public final static boolean DEBUG_BAM_SPLITTER = false;
 
+	static final String BOUNDED_TRAVERSAL_PROPERTY = "hadoopbam.bam.bounded-traversal";
```
heuermh: Should this be `public static` as `INTERVALS_PROPERTY` is?  Also needs javadoc.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -72,31 +74,61 @@
...
+
 	/**
 	 * Filter by region, like <code>-L</code> in SAMtools. Takes a comma-separated
 	 * list of intervals, e.g. <code>chr1:1-20000,chr2:12000-20000</code>. For
 	 * programmatic use {@link #setIntervals(Configuration, List)} should be preferred.
 	 */
 	public static final String INTERVALS_PROPERTY = "hadoopbam.bam.intervals";
 
+	static final String TRAVERSE_UNPLACED_UNMAPPED_PROPERTY = "hadoopbam.bam.traverse-unplaced-unmapped";
```
heuermh: Should this be `public static` as `INTERVALS_PROPERTY` is?  Also needs javadoc.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -72,31 +74,61 @@
...
-		StringBuilder sb = new StringBuilder();
-		for (Iterator<T> it = intervals.iterator(); it.hasNext(); ) {
-			Locatable l = it.next();
-			sb.append(String.format("%s:%d-%d", l.getContig(), l.getStart(), l.getEnd()));
-			if (it.hasNext()) {
-				sb.append(",");
+		setTraversalParameters(conf, intervals, false);
+	}
+
+	public static <T extends Locatable> void setTraversalParameters(Configuration conf,
```
heuermh: Needs javadoc.

src/test/java/org/seqdoop/hadoop_bam/TestBAMInputFormat.java
```diff
@@ -26,11 +26,12 @@
   private TaskAttemptContext taskAttemptContext;
   private JobContext jobContext;
   
-  private void completeSetup(List<Interval> intervals) {
+  private void completeSetup(boolean boundedTraversal, List<Interval> intervals, boolean
```
heuermh: I know this is just test code, but with two boolean parameters it is not really clear in the test methods below which situation is under test.

How about adding new methods `setupBoundedTraversalUnplacedUnmapped(List<Interval>)` and similar that delegate to this one?

tomwhite:
@heuermh thanks for the review. I have addressed all your points: I added documentation to all methods, and also entries in the README to cover bounded traversal. I made the test code clearer, and also added a negative test for the case of mapped only reads (which are not currently supported).

Please let me know if it looks OK now.

PR #140, opened 2017-08-01
Add codepath to calculate splits from BAM index.

fnothaft:
This is pretty ugly right now (see debugging statements, commented out code, etc), but this code adds a path to identify splits from a BAM file using the BAM index on disk. It does this by using the linear index that accompanies the BAM file to identify the first block in the linear index that starts after the start/end of a suggested split. This code works (for a single indexed file, haven't tested multiple files yet) and appears to show modest performance improvements in split discovery.

As an aside, while the `htsjdk.samtools.LinearIndex` class is public, there doesn't appear to be a public way to access a `LinearIndex`, so I've added the cruddy `LinearBAMIndex` class.

As said earlier, this code is very WIP, but I wanted to push it early to get general comments on the strategy, and to start a conversation about where this might fit in, relative to the splitting BAI code.

### DIFF
--- a/src/main/java/htsjdk/samtools/LinearBAMIndex.java
+++ b/src/main/java/htsjdk/samtools/LinearBAMIndex.java
@@ -1,0 +1,39 @@
+// Copyright (c) 2017
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to
+// deal in the Software without restriction, including without limitation the
+// rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+// sell copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+// IN THE SOFTWARE.
+package htsjdk.samtools;
+
+import htsjdk.samtools.CachingBAMFileIndex;
+import htsjdk.samtools.LinearIndex;
+import htsjdk.samtools.SAMSequenceDictionary;
+import htsjdk.samtools.seekablestream.SeekableStream;
+
+/**
+ * The htsjdk APIs for accessing the linear BAM index are private...
+ */
+public class LinearBAMIndex extends CachingBAMFileIndex {
+
+        public LinearBAMIndex(SeekableStream stream, SAMSequenceDictionary dict) {
+                super(stream, dict);
+        }
+        
+        public LinearIndex getLinearIndex(int idx) {
+                return getQueryResults(idx).getLinearIndex();
+        }
+}
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
@@ -28,9 +28,13 @@ import htsjdk.samtools.BAMFileReader;
 import htsjdk.samtools.BAMFileSpan;
 import htsjdk.samtools.BAMIndex;
 import htsjdk.samtools.Chunk;
+import htsjdk.samtools.LinearBAMIndex;
+import htsjdk.samtools.LinearIndex;
 import htsjdk.samtools.QueryInterval;
 import htsjdk.samtools.SAMFileHeader;
+import htsjdk.samtools.SAMFileSpan;
 import htsjdk.samtools.SAMSequenceDictionary;
+import htsjdk.samtools.SamInputResource;
 import htsjdk.samtools.SamReader;
 import htsjdk.samtools.SamReaderFactory;
 import htsjdk.samtools.util.Interval;
@@ -56,6 +60,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 
+import htsjdk.samtools.BAMIndex;
 import htsjdk.samtools.seekablestream.SeekableStream;
 
 import org.apache.hadoop.conf.Configuration;
@@ -85,6 +90,14 @@ public class BAMInputFormat
 	public static final String BOUNDED_TRAVERSAL_PROPERTY = "hadoopbam.bam.bounded-traversal";
 
 	/**
+	 * If set to true, enables the use of BAM indices to calculate splits.
+	 * For programmatic use
+	 * {@link #setEnableBAISplitCalculator(Configuration, boolean)} should be preferred.
+         * By default, this split calculator is disabled in favor of the splitting-bai calculator.
+	 */
+	public static final String ENABLE_BAI_SPLIT_CALCULATOR = "hadoopbam.bam.enable-bai-splitter";
+    
+	/**
 	 * Filter by region, like <code>-L</code> in SAMtools. Takes a comma-separated
 	 * list of intervals, e.g. <code>chr1:1-20000,chr2:12000-20000</code>. For
 	 * programmatic use {@link #setIntervals(Configuration, List)} should be preferred.
@@ -109,6 +122,14 @@ public class BAMInputFormat
 			List<T> intervals) {
 		setTraversalParameters(conf, intervals, false);
 	}
+
+        /**
+         * Enables or disables the split calculator that uses the BAM index to calculate splits.
+         */
+        public static void setEnableBAISplitCalculator(Configuration conf,
+                        boolean setEnabled) {
+            conf.setBoolean(ENABLE_BAI_SPLIT_CALCULATOR, setEnabled);
+        }
 
 	/**
 	 * Only include reads that overlap the given intervals (if specified) and unplaced
@@ -186,7 +207,13 @@ public class BAMInputFormat
 		return splits.stream()
 				.filter(split -> !((FileSplit) split).getPath().getName().endsWith(
 						SplittingBAMIndexer.OUTPUT_FILE_EXTENSION))
+                                .filter(split -> !((FileSplit) split).getPath().getName().endsWith(
+                                                BAMIndex.BAMIndexSuffix))
 				.collect(Collectors.toList());
+        }
+    
+    	static Path getBAIPath(Path path) {
+		return path.suffix(BAMIndex.BAMIndexSuffix);
 	}
 
 	/** Returns a {@link BAMRecordReader} initialized with the parameters. */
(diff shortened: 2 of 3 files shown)


ryan-williams:
Haven't looked at this impl, but I gave up on an attempt to do this because a `.bai` gives no offsets into blocks of unmapped reads, and I saw BAMs where there were more than 32MB of unmapped reads, meaning I would have had to drop down to split-guessing either way. 

Interested in whether I missed something there or you have thoughts about that.

fnothaft:
@ryan-williams yeah, I haven't handled the unmapped reads case in an elegant way. If we have a .bai, that means that the reads are sorted and thus all unmapped reads are at the end of the file. I generate one split that is from the last bin offset to the end of the file, so all the unmapped reads go there. This is obviously not a good solution, but it works for a first hack.

ryan-williams:
> If we have a .bai, that means that the reads are sorted and thus all unmapped reads are at the end of the file.

IIRC, there are blocks of "placed" unmapped reads at the end of each contig's portion of the BAM, and the case that stymied me was somewhere in the middle of an indexed BAM.

fnothaft:
That's in violation of the SAM spec, no?

> Sorting order of alignments. Valid values: unknown (default), unsorted, queryname and coordinate. For coordinate sort, the major sort key is the RNAME field, with order defined by the order of @SQ lines in the header. The minor sort key is the POS field. For alignments with equal RNAME and POS, order is arbitrary. All alignments with ‘*’ in RNAME field follow alignments with some other value but otherwise are in arbitrary order.

Or, are you saying that `RNAME` is defined but the read flags show unmapped and `POS` is 0?

ryan-williams:
`RNAME` and `POS` can both be set though the read is unmapped (indicated by the unmapped bit).

> RNAME: Reference sequence NAME of the alignment. If @SQ header lines are present, RNAME (if not `*`) must be present in one of the SQ-SN tag. An unmapped segment without coordinate has a `*` at this field. **However, an unmapped segment may also have an ordinary coordinate such that it can be placed at a desired position after sorting**. If RNAME is `*`, no assumptions can be made about POS and CIGAR.

emphasis mine

Later, in the BAI section:



`unmapped_beg` and `unmapped_end` bookend this kind of block of _{placed, unmapped}_ reads, but the BAI gives you no offsets into the middle of such a section, afaict.

fnothaft:
Yeah, the placed unmapped isn't news to me. The part that is news to me is that these reads should be sorting into a contiguous block at the end of the contig?

ryan-williams:
Yea, I'm having trouble reconciling two mutually exclusive things:

1. reads are sorted by `(RNAME, POS)`, even if unmapped
2. placed+unmapped reads are in a contiguous block at the end of each contig's BAI section

The `unmapped_{beg,end}` info above seems to imply 2., as does my experience where I gave up on BAI-informed-split-inference, but that seems to contradict 1.…

Separately / either way, you can end up with <4096bp having >64MB of BAM, in which case you also have to split-"guess", right?

fnothaft:
Yeah, I'm having the same confusion.

fnothaft:
@ryan-williams my latest commit now guesses splits once we get to unmapped blocks.

tomwhite:
@fnothaft I think this would be useful - thanks for working on it! I think this will need testing with a variety of BAM files of different sizes.

BTW you've probably seen it, but I've been working on (unplaced) unmapped reads here: https://github.com/HadoopGenomics/Hadoop-BAM/pull/136, although that is for the guessing case, so should be orthogonal to your change.

fnothaft:
Thanks @tomwhite! I've been using this on a variety of BAM files both in HDFS and in S3.

fnothaft:
Hi @tomwhite! I've made the cleanup we were discussing over email; let me know if you have any further suggestions on this. I've retested on a small set of BAMs with the code enabled and disabled and all seems OK.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -230,12 +261,16 @@ public int compare(InputSplit a, InputSplit b) {
...
 		for (int i = 0; i < origSplits.size();) {
 			try {
-				i = addIndexedSplits      (origSplits, i, newSplits, cfg);
-			} catch (IOException e) {
-				i = addProbabilisticSplits(origSplits, i, newSplits, cfg);
+                                if (cfg.getBoolean(ENABLE_BAI_SPLIT_CALCULATOR, false)) {
+                                        i = addBAISplits    (origSplits, i, newSplits, cfg);
+                                } else {
+				        i = addIndexedSplits(origSplits, i, newSplits, cfg);
+                                }
```
tomwhite: I don't think this is quite right. Even if `hadoopbam.bam.enable-bai-splitter` is true, if there is a splitting-bai file, then it should be used, since it has been generated for finding splits.

So I think the precedence order should be 1) use splitting-bai (if it exists), 2) using bai (if it exists and `hadoopbam.bam.enable-bai-splitter` is true), 3) use heuristics. If 1 or 2 fail, then fall back to 3 also.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -230,12 +261,16 @@ public int compare(InputSplit a, InputSplit b) {
...
+                        } catch (IOException e) {
+				i = addProbabilisticSplits  (origSplits, i, newSplits, cfg);
 			}
 		}
-		return filterByInterval(newSplits, cfg);
+                return filterByInterval(newSplits, cfg);
```
tomwhite: Nit: whitespace

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -84,6 +88,14 @@
 	public static final String BOUNDED_TRAVERSAL_PROPERTY = "hadoopbam.bam.bounded-traversal";
 
 	/**
+	 * If set to true, enables the use of BAM indices to calculate splits.
+	 * For programmatic use
+	 * {@link #setBAISplitCalculatorEnable(Configuration, boolean)} should be preferred.
+         * By default, this split calculator is disabled in favor of the splitting-bai calculator.
+	 */
```
tomwhite: This should be updated to reflect the precedence I suggested in another comment.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -109,6 +121,19 @@
 		setTraversalParameters(conf, intervals, false);
 	}
 
+        /**
+         * Enables or disables the split calculator that uses the BAM index to calculate splits.
+         */
+        public static void setBAISplitCalculatorEnable(Configuration conf,
```
tomwhite: setEnableBAISplitCalculator?

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -109,6 +121,19 @@
...
+                        boolean setEnabled) {
+            conf.setBoolean(ENABLE_BAI_SPLIT_CALCULATOR, setEnabled);
+        }
+
+        /**
+         * The file extension corresponding to a BAM index.
+         */
+        private static final String INDEX_FILE_EXTENSION = ".bai";
```
tomwhite: Use `htsjdk.samtools.BAMIndex.BAMIndexSuffix`.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -185,7 +210,13 @@ static Path getIdxPath(Path path) {
 		return splits.stream()
 				.filter(split -> !((FileSplit) split).getPath().getName().endsWith(
 						SplittingBAMIndexer.OUTPUT_FILE_EXTENSION))
+                                .filter(split -> !((FileSplit) split).getPath().getName().endsWith(
+                                                INDEX_FILE_EXTENSION))
 				.collect(Collectors.toList());
+        }
+    
+    	static Path getBAIPath(Path path) {
+		return path.suffix(".bai");
```
tomwhite: Use `htsjdk.samtools.BAMIndex.BAMIndexSuffix`.

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -295,7 +330,127 @@ private int addIndexedSplits(
 		}
 		return splitsEnd;
 	}
-
+        
+        private int addBAISplits(List<InputSplit> splits,
```
tomwhite: This method needs some comments to explain what is going on.

src/main/java/htsjdk/samtools/LinearBAMIndex.java
```diff
@@ -0,0 +1,39 @@
...
+
+import htsjdk.samtools.CachingBAMFileIndex;
+import htsjdk.samtools.LinearIndex;
+import htsjdk.samtools.SAMSequenceDictionary;
+import htsjdk.samtools.seekablestream.SeekableStream;
+
+/**
+ * The htsjdk APIs for accessing the linear BAM index are private...
+ */
+public class LinearBAMIndex extends CachingBAMFileIndex {
```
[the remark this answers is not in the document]
fnothaft: Hi @tomwhite! I've made this public because it is located in `htsjdk.samtools` and needs to be accessed from `org.seqdoop.hadoop_bam`. The class has to be in `htsjdk.samtools` because it needs access to the package-private `CachingBAMFileIndex` class.

fnothaft:
Thanks for the review @tomwhite! I'll address your comments and turn this around today.

src/main/java/htsjdk/samtools/LinearBAMIndex.java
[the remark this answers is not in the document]
tomwhite: Oh yes, this is in a different package so it needs to be public. Can you open an htsjdk issue to increase its visibility please?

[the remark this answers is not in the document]
fnothaft: Definitely, will do.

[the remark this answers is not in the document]
fnothaft: Opened https://github.com/samtools/htsjdk/issues/975.

fnothaft:
Hi @tomwhite! I believe this is good to go from my end.

src/test/java/org/seqdoop/hadoop_bam/TestBAMInputFormat.java
```diff
@@ -80,6 +80,27 @@ public void testMultipleSplits() throws Exception {
   }
 
   @Test
+  public void testMultipleSplitsBaiEnabled() throws Exception {
```
tomwhite: This test doesn't test the addBAISplits codepath, it just tests that when there is no .bai file and the bai split calculator is enabled it falls back to finding probabilistic splits.

Can you add a positive test for the .bai file case?

src/main/java/org/seqdoop/hadoop_bam/BAMInputFormat.java
```diff
@@ -186,7 +207,13 @@ static Path getIdxPath(Path path) {
 		return splits.stream()
 				.filter(split -> !((FileSplit) split).getPath().getName().endsWith(
 						SplittingBAMIndexer.OUTPUT_FILE_EXTENSION))
+                                .filter(split -> !((FileSplit) split).getPath().getName().endsWith(
+                                                BAMIndex.BAMIndexSuffix))
 				.collect(Collectors.toList());
+        }
+    
+    	static Path getBAIPath(Path path) {
+		return path.suffix(BAMIndex.BAMIndexSuffix);
```
tomwhite: We also need to cover the case for files named `file.bai` as well as `file.bam.bai`.

fnothaft:
Hi @tomwhite. Sorry for the confusion with the tests, I'd set the sort order incorrectly. This should be fixed now, along with your .bam.bai/.bai comment. I've squashed this down; let me know if you'd like any other changes.

tomwhite:
+1

fnothaft:
Thanks @tomwhite!

PR #142, opened 2017-08-18
Eliminate buffer copy bug in BGZFSplitGuesser.

fnothaft:
In the existing BGZFSplitGuesser code, data from the InputStream is copied into a byte array, which is used as a buffer for finding the next BGZF block start. Previously, this code assumed that a non-negative return from the read call meant that the full amount of bytes requested were copied. This does not always happen, which can break the split guessing code. This patch adds logic to make sure that the buffer is filled.

Additionally, this patch rewrites the BGZF header guessing code. I did this as a debugging process (i.e., to better understand where the error was occurring) and I don't think it makes a big impact on correctness. It does seem to be slightly more performant; the ByteBuffer copy seems to be pretty slow. @tomwhite If you'd rather not rewrite that function, let me know, and I can take it out for later.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/util/BGZFSplitGuesser.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/util/BGZFSplitGuesser.java
@@ -39,10 +39,14 @@ public class BGZFSplitGuesser {
 	private Seekable seekableInFile;
 	private       ByteArraySeekableStream in;
 	private final ByteBuffer buf;
-
-	private final static int BGZF_MAGIC     = 0x04088b1f;
-	private final static int BGZF_MAGIC_SUB = 0x00024342;
-	private final static int BGZF_SUB_SIZE  = 4 + 2;
+    private String name = null;
+    
+    private final static int BGZF_MAGIC_0 = 0x1f;
+    private final static int BGZF_MAGIC_1 = 0x8b;
+    private final static int BGZF_MAGIC_2 = 0x08;
+    private final static int BGZF_MAGIC_3 = 0x04;
+    private final static int BGZF_MAGIC = 0x04088b1f;
+    private final static int BGZF_MAGIC_SUB = 0x00024342;
 
 	public BGZFSplitGuesser(InputStream is) {
 		inFile = is;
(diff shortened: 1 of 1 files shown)


src/main/java/org/seqdoop/hadoop_bam/util/BGZFSplitGuesser.java
```diff
@@ -69,15 +77,24 @@ public long guessNextBGZFBlockStart(long beg, long end)
 		// the previous one, we need 0xfffe bytes for the start, and then 0xffff
 		// for the block we're looking for.
 
-		byte[] arr = new byte[2*0xffff - 1];
+		byte[] arr = new byte[0xffff];
```
ryan-williams: curious how this change interacts w the comment above? at the very least the latter seems outdated now?

tomwhite:
@fnothaft I noticed that this logic is somewhat duplicated with BAMSplitGuesser. In fact, the same thing has been fixed there: https://github.com/HadoopGenomics/Hadoop-BAM/blob/master/src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java#L126-L137. So you could use the same approach in BGZFSplitGuesser.

Do you have a way of testing this?

There are actually three split guessers: BAMSplitGuesser, BCFSplitGuesser, and BGZFSplitGuesser. The first two subclass BaseSplitGuesser so they can share some code, so maybe BGZFSplitGuesser could subclass it too, to share the implementation of `guessNextBGZFPos`. That's slightly different to this issue though, so should probably be done separately.

I would break out the changes to the BGZF header guessing to a separate PR. That should cover the other split guessers too.

fnothaft:
Hi @tomwhite! Indeed, I was noticing that there are 3 split guessers as I was testing and working on this. I have been testing this on a mix of BGZF'ed FASTQs and VCFs. If you'd prefer, I can make `BGZFSplitGuesser` extend `BaseSplitGuesser` and then redo this using the logic from https://github.com/HadoopGenomics/Hadoop-BAM/blob/master/src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java#L126-L137. The changes in the BGZF header detection code were more for debugging than anything, but they do appear to improve performance. I'm thinking I'll split the BaseSplitGuesser and array copy fix into a separate PR, and then we can review the BGZF header detection changes later. I'd love to get the array copy bug fixed in 7.9.0.

tomwhite:
@fnothaft sounds good. Do you want to close or rename this PR now that #149 is in?

fnothaft:
I'll close this for now. Thanks @tomwhite!

PR #146, opened 2017-08-24
Use slf4j logging instead of System.err.println

### LINKED ISSUE #145: Use slf4j logging instead of System.err.println

heuermh:
Fixes #145

### DIFF
--- a/src/test/resources/log4j.properties
+++ b/src/test/resources/log4j.properties
@@ -1,5 +1,7 @@
-log4j.rootLogger = INFO, out
+log4j.rootLogger = WARN, out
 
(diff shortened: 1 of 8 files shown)


src/test/resources/log4j.properties
```diff
@@ -1,5 +1,7 @@
-log4j.rootLogger = INFO, out
+log4j.rootLogger = ERROR, out
```
fnothaft: I'd suggest WARN instead of ERROR, but +1 on moving to a higher level than INFO.

heuermh:
Addressed review comments, using WARN instead of error.

heuermh:
Thanks!

PR #153, opened 2017-09-11
Reformat source code for consistent use of whitespace.

### LINKED ISSUE #147: Reformat source code for consistent use of whitespace
At a minimum we'd have to agree on:

Use tab characters or spaces  (spaces)
Tab/indent size  (2 or 4)
Space before parens  (1)
Space around operators  (1)
Space before left brace  (1)
Space before keywords  (1)
Space in ternary operator  (1)
Space after comma in type arguments  (1)
Space after comma  (1)
Space after type cast  (1)
Else on new line  (Yes)
Right margin  (120 columns)

My preferences in parens.  I'd suggest doing this with a source code formatter all in one commit after the next release.

heuermh:
Fixes #147.

I created a code style for Intellij IDEA (checked in at `etc/IntelliJ IDEA Code Style.xml`) and ran Reformat Code.  All source code changes should be whitespace and braces only.

### DIFF
--- a/etc/IntelliJ IDEA Code Style.xml
+++ b/etc/IntelliJ IDEA Code Style.xml
@@ -1,0 +1,32 @@
+<code_scheme name="Default (1)">
+  <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
+  <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
+  <option name="IMPORT_LAYOUT_TABLE">
+    <value>
+      <package name="java" withSubpackages="true" static="false" />
+      <package name="javax" withSubpackages="true" static="false" />
+      <package name="" withSubpackages="true" static="false" />
+      <emptyLine />
+      <emptyLine />
+      <package name="" withSubpackages="true" static="true" />
+    </value>
+  </option>
+  <option name="JD_ALIGN_PARAM_COMMENTS" value="false" />
+  <option name="JD_ALIGN_EXCEPTION_COMMENTS" value="false" />
+  <ScalaCodeStyleSettings>
+    <option name="USE_SCALADOC2_FORMATTING" value="false" />
+    <option name="ENABLE_SCALADOC_FORMATTING" value="false" />
+  </ScalaCodeStyleSettings>
+  <codeStyleSettings language="JAVA">
+    <option name="ELSE_ON_NEW_LINE" value="true" />
+    <option name="WHILE_ON_NEW_LINE" value="true" />
+    <option name="CATCH_ON_NEW_LINE" value="true" />
+    <option name="FINALLY_ON_NEW_LINE" value="true" />
+    <option name="SPACE_WITHIN_ARRAY_INITIALIZER_BRACES" value="true" />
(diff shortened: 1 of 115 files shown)


fnothaft:
+1!

tomwhite:
Thanks for doing this @heuermh!

The code style and particularly the whitespace is particularly inconsistent in this codebase, so I'm +1 on fixing it. I think we should do it in a major release, though, since it's quite a disruptive change.

Also, I think we should enforce the style by using checkstyle or similar. I started writing a PR a while ago to do this, it's incomplete, but you can see the idea here: https://github.com/HadoopGenomics/Hadoop-BAM/compare/master...tomwhite:checkstyle?expand=1

What do you think?

heuermh:
> Also, I think we should enforce the style by using checkstyle or similar.

I agree, and as an example have it configured here
https://github.com/heuermh/dishevelled/tree/master/compress/src/build/conf
https://github.com/heuermh/dishevelled/blob/master/parent/pom.xml#L315

We would need to sync the configuration in the source code formatter with that enforced by checkstyle.

ADAM formats source on the fly as part of the build, I think that might be more trouble than it is worth.

tomwhite:
Great!

> ADAM formats source on the fly as part of the build, I think that might be more trouble than it is worth.

Agreed.

fnothaft:
+1

tomwhite:
Rather than choosing our own formatting preferences, I would rather choose an existing style guide to use. The [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html) looks like the one that is most used these days (the old Sun one has not been updated since 1999).

At first I thought we should use Checkstyle to enforce the style, but after playing with the [Google checkstyle](http://checkstyle.sourceforge.net/google_style.html) settings I found that they produce lots (hundreds) of warnings even for source code that has been formatted to adhere to the Google Java Style Guide. For example, it produces lots of warnings about there not being blank lines between import groups even though the guide says there shouldn't be blank lines.

So I think it would be preferable to use a tool that checks if the source is formatted correctly according to the Google Java Style Guide, and fail the build if it's not. This is such a tool: https://github.com/coveo/fmt-maven-plugin, and it's easy to integrate with the Maven build. The source can also be manually reformatted using `mvn fmt:format` if needed.

I've produce a  patch to do this here: #182. Please take a look!

heuermh:
The Google Java Style Guide looks fairly reasonable to me.

The only thing I don't like is the [missing newline after `}`](https://google.github.io/styleguide/javaguide.html#s4.1-braces), I prefer
```java
return () -> {
  while (condition()) {
    method();
  }
};

return new MyClass() {
  @Override public void method() {
    if (condition()) {
      try {
        something();
      }
      catch (ProblemException e) {
        recover();
      }
    }
    else if (otherCondition()) {
      somethingElse();
    }
    else {
      lastThing();
    }
  }
};
```

to
```java
return () -> {
  while (condition()) {
    method();
  }
};

return new MyClass() {
  @Override public void method() {
    if (condition()) {
      try {
        something();
      } catch (ProblemException e) {
        recover();
      }
    } else if (otherCondition()) {
      somethingElse();
    } else {
      lastThing();
    }
  }
};
```

I could of course could be outvoted.  :)

fnothaft:
+1 to using the Google Java style guide.

heuermh:
Closing as WontFix

PR #157, opened 2017-09-22
Support specifying validation stringency in VCFRecordReader.

### LINKED ISSUE #156: Allow bad VCF rows to be waived with validation stringency
We have some large VCFs with a few malformed rows. We'd like to waive the failing rows by using a validation stringency check.

fnothaft:
Resolves #156. Wraps the VCFCodec decode function in a try-catch that uses the user specified validation stringency to throw, log, or skip errors.


fnothaft:
Still needs unit tests; will be circling back to confirm that this is validated.

src/main/java/org/seqdoop/hadoop_bam/VCFRecordReader.java
```diff
@@ -140,7 +173,21 @@
 				}
 			}
 
-			final VariantContext v = codec.decode(line);
+                        final VariantContext v;
+                        try {
+				v = codec.decode(line);
+			} catch (Exception e) {
```
cmnbroad: @fnothaft Can you add any more detail about what kinds of VCF failures you're trying to recover from ? Masking any Exception thrown during parsing seems like much too coarse grained of a knob (we're trying to remove instances of this pattern from htsjdk). You should be seeing TribbleExceptions on VCF parsing errors At a minimum, narrowing the catch to TribbleException would be better, but even that seems coarse grained. If parsing errors are manifesting in some other way, I'd like to understand the root cause and see if we can fix that.

cmnbroad: One other minor thing, in htsjdk, ValidationStringency is a SAM thing used by the samtools package; its not used directly in the Tribble/variant packages, so it might be confusing to overload that term for this purpose.

heuermh: Not speaking for @fnothaft, in that I don't know what specifically lead to this PR.

Generally we see two classes of issues, where htsjdk implements or interprets the VCF specification incorrectly, and where data are invalid according to the VCF specification yet the user still wants to work with them. A good example of the latter is `Number=1,Type=String` instead of `Number=1,Type=Integer` for `##FORMAT=<ID=PS`, which is found often in public data sets.

We've abused `ValidationStringency` somewhat in ADAM by using it in many places for similar behavior.  I've been uncomfortable with that, so I wonder if a new tiny API with a new enum and static methods for the if strict throw else if lenient log warning pattern might be useful.  We have an issue around that at https://github.com/bigdatagenomics/adam/issues/1572.  If htsjdk ever comes around to slf4j, I suppose it could live there.  Otherwise I could push something to Maven Central under an Apache 2 license rather quickly.

fnothaft: Hi @cmnbroad! My apologies for the slow reply, lots of travel/etc recently. We've got users who have large VCF datasets (e.g., >10k VCFs) where there are a small subset of records that fail validation. I'd have to go back to my notes, but I'm not 100% sure whether the VCF lines that fail validation are out of line with the VCF spec, or whether they're lines that are OK in the VCF 4.3 spec (but not supported in 4.2 --> I think this is what @heuermh meant when he said that htsjdk misinterpreted the VCF spec?).

I'll go ahead and make the exception narrower, that seems reasonable to me. Additionally, I'm going to make sure that we log the failed line even on STRICT. IIRC, the exceptions we've been seeing don't log the failing line, which makes the issue hard to debug. That said, without getting you the actual exception, that's not very useful feedback. Let me see if I can track down the specific exception we ran into.

src/main/java/org/seqdoop/hadoop_bam/VCFRecordReader.java
```diff
@@ -140,7 +174,26 @@
 				}
 			}
 
-			final VariantContext v = codec.decode(line);
+                        final VariantContext v;
+                        try {
+				v = codec.decode(line);
+			} catch (TribbleException e) {
+				if (stringency == ValidationStringency.STRICT) {
+					if (stringency == ValidationStringency.LENIENT &&
```
heuermh: If strict, we're throwing `e`, so why log here?  Shouldn't that be downstream's responsibility?

src/main/java/org/seqdoop/hadoop_bam/VCFRecordReader.java
```diff
@@ -140,7 +174,26 @@
...
+                                            logger.isErrorEnabled()) {
+						logger.error("Parsing line {} failed with {}. Skipping...",
```
heuermh: We're not skipping here

src/main/java/org/seqdoop/hadoop_bam/VCFRecordReader.java
fnothaft: htsjdk doesn't include the offending line in all exception messages IIRC; logging an error message with the line and exception makes debugging the exception much easier.

fnothaft:
Rebased and addressed reviewer comments.

tomwhite:
This looks reasonable to me now. I'll commit this in a day or so unless there are any more comments.

cmnbroad:
@fnothaft I think the change look fine, but can you add a test ?

heuermh:
Looks good to me with the unit tests.  Ping for merge?

src/test/java/org/seqdoop/hadoop_bam/TestVCFInputFormat.java
```diff
@@ -100,6 +102,10 @@ public void setup() throws IOException, NoSuchMethodException, IllegalAccessExce
             BGZFCodec.class.getCanonicalName());
         conf.setInt(FileInputFormat.SPLIT_MAXSIZE, 100 * 1024); // 100K
 
+        if (filename == "small_snpeff.vcf") {
```
cmnbroad: Thanks for adding the test case to cover this - it will help keep test coverage up as functionality is added. However, I think its confusing to embed exceptional code paths throughout the file like this. It shouldn't be that hard to add a separate test case for this change; but at the very least I'd request that you include comments describing why this code is here. And, since presumably the test file is invalid (white space ?), it would be useful to name it something that indicates that. Absent those things, I don't think anyone coming along after the fact will be able make any sense out of this.

cmnbroad: I would avoid using reference equality operators on strings - it happens to work in this case due to string folding, but these comparisons should really use .equals instead of == or !=.

tomwhite: @cmnbroad @fnothaft @heuermh I've made the test standalone to avoid these issues. Hope that's acceptable to everyone!

fnothaft:
+1, sorry I haven't been able to circle back on the last change request, been slammed this week. Thanks @tomwhite!

PR #163, opened 2017-09-27
In NIOFileUtil.deleteRecursive only delete directory if it exists. 

### LINKED ISSUE #131: NIOFileUtil.deleteRecursive delete if exists
Hi, 

I am using an Amazon S3 NIO library and seeing a small issue in the `NIOFileUtil.deleteRecursive` function. S3 has no concept of directories, so once all of the files have been deleted, the parent folder not longer exists. This causes the final `postVisitDirectory()` method to throw a `NoSuchFile` exception. Would it be reasonable to change `deleteRecursive()` to use `Files.deleteIfExists()` inside `postVisitDirectory()` like so? 

```
  static void deleteRecursive(Path directory) throws IOException {
    Files.walkFileTree(directory, new SimpleFileVisitor<Path>() {
      @Override
      public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
        Files.delete(file);
        return FileVisitResult.CONTINUE;
      }
      @Override
      public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOException {
        Files.deleteIfExists(dir);
        return FileVisitResult.CONTINUE;
      }
    });
  }
```

Thanks,
David

tomwhite:
Fixes #131.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/util/NIOFileUtil.java
@@ -68,7 +68,7 @@ public class NIOFileUtil {
       }
       @Override
       public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOException {
-        Files.delete(dir);
+        Files.deleteIfExists(dir);
         return FileVisitResult.CONTINUE;
       }
     });


tomwhite:
Thanks for taking a look @cmnbroad

Issue #33, opened 2015-11-19
Add extra check to finding split point via guessing (if profiling says it's fine)


davidadamsphd:
Implement the extra validation if this ticket: https://github.com/HadoopGenomics/Hadoop-BAM/issues/32 goes well

tomwhite:
Fixed in #164

Issue #32, opened 2015-11-19
Profile adding extra checks to "guessing" which record to decode


davidadamsphd:
It's unclear of the performance cost of decoding the rest of the (first three?) records when trying to find a start point. This should be benchmarked on a large BAM ~100-300GB against current behavior.

tomwhite:
Fixed in #164

PR #164, opened 2017-11-01
Decode BAM fields eagerly in BAMSplitGuesser when verifying 3 records…

tomwhite:
… once a candidate position has been found. This strengthens the check from the
previous one, which just checked that the CIGAR string could be decoded.

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java
@@ -23,9 +23,10 @@
 package org.seqdoop.hadoop_bam;
 
 import htsjdk.samtools.BAMFileSpan;
+import htsjdk.samtools.SAMFileHeader;
 import htsjdk.samtools.SAMFileSpan;
 import htsjdk.samtools.SAMRecord;
-import htsjdk.samtools.SAMUtils;
+import htsjdk.samtools.SAMRecordHelper;
 import htsjdk.samtools.SamInputResource;
 import htsjdk.samtools.SamReader;
 import htsjdk.samtools.SamReaderFactory;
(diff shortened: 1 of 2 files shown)


tomwhite:
This addresses #33 and #32. (I ran this on a large - 10s of GB - BAM, and didn't see any noticeable impact on performance.)

tomwhite:
There are no unit tests since the BAM file that exhibits the problem is too large to check in, and I don't know how to create one that fails without the fixes.

fnothaft:
CCing @heuermh and @ryan-williams, both of whom have BAM files that cause the split picker to fail. They may have BAMs that have an error in an early split?

heuermh:
I ran into two distinct errors based on bad splits (https://github.com/bigdatagenomics/adam/issues/1656 and https://github.com/bigdatagenomics/adam/issues/1657), and can make those BAM files available on the bdg cluster for analysis.  Is there a quick way to test this fix, or should I just bundle up into ADAM and try that way?

fnothaft:
> Is there a quick way to test this fix, or should I just bundle up into ADAM and try that way?

I would probably just do that.

src/main/java/org/seqdoop/hadoop_bam/BAMSplitGuesser.java
```diff
@@ -186,7 +187,8 @@ public long guessNextBAMRecordStart(long beg, long end)
 						if (record == null) {
 							break;
 						}
-						record.getCigar(); // force decoding of CIGAR
+						record.setHeader(header);
```
cmnbroad: One suggestion to make this more robust would be to use `setHeaderStrict` instead of `setHeader`. `setHeader` is tolerant of a record with a reference index or mate reference index that can't be resolved against the header's sequence dictionary. `setHeaderStrict` is identical, but throws `IllegalArgumentException` if either one can't be resolved against the header. So in that sense its a "safe" way to quickly reject a bad split guess before calling eagerDecode.

tomwhite: @cmnbroad good idea! I've made this change.

tomwhite:
@heuermh it would be easiest if you could use this patch to test on your BAM files. Thanks!

heuermh:
To follow up, this fix works for me, solving those issues mentioned above.

Issue #109, opened 2016-07-01
Provide a way to specify a custom Deflater


akiezun:
htsjdk allows for specifying custom deflaters for writing bam files. We (GATK) would like to use the `IntelDeflater` from the Intel GKL https://github.com/Intel-HLS/GKL

This can be done by setting ` BlockCompressedOutputStream.setDefaultDeflaterFactory(new IntelDeflaterFactory());` or by setting the right deflater on the writers

PR #167, opened 2017-11-07
Add options to use the Intel inflater or deflater when reading or

tomwhite:
writing BAM files. Use of this option requires that you provide
the GKL dependency (https://github.com/Intel-HLS/GKL) on the classpath.


tomwhite:
Addresses #109

src/main/java/org/seqdoop/hadoop_bam/IntelGKLAccessor.java
```diff
@@ -0,0 +1,20 @@
+package org.seqdoop.hadoop_bam;
```
heuermh: needs a license header

tomwhite: Fixed

PR #176, opened 2017-12-15
Specifies correct extensions for various file formats

### LINKED ISSUE #175: OutputFormats should write out the correct extension for their file type
E.g., VCFOutputFormat just writes out the compressor extension. This makes it impossible to load sharded output from tools that autodetect the format from the file extensions.

fnothaft:
Resolves #175. Would love to take this in the 7.9.1 release if possible.


src/main/java/org/seqdoop/hadoop_bam/KeyIgnoringVCFOutputFormat.java
```diff
@@ -97,7 +97,7 @@ public void readHeaderFrom(SeekableStream in) throws IOException {
 		Configuration conf = ctx.getConfiguration();
 		boolean isCompressed = getCompressOutput(ctx);
 		CompressionCodec codec = null;
-		String extension = "";
+		String extension = ".vcf";
```
tomwhite: Won't this get replaced by the compressor extension in the next code block anyway? So it will be "file.bgz" not "file.vcf.bgz".

fnothaft:
@tomwhite SGTM! We can override the behavior downstream in ADAM, so don't hold up 7.9.1 for this patch.

PR #197, opened 2018-03-15
delete shards as they're merged instead of afterwards

lbergelson:
this should reduce the total amount of scratch disk space that is needed for performing merge operations and should have no effect on performance


lbergelson:
@tomwhite Could you take a look at this when you get a chance?  
@jamesemery This should help the space issues a bit.

Issue #199, opened 2018-06-06
Reading sharded SAM / CRAM fails on some filesystems


lbergelson:
It seems like reading sharded sam / cram files correctly depends on what file system you're using.  Particularly, the order the shards are read seems to be different depending on which underlying filesystem you're using.  I suspect that there is a call to get an iterator over the part files that gets an iterator directly from the file system and then doesn't sort it.

`BAMInputFormat` seems to work because it overrides `getSplits` and then sorts them based on the returned path.  Extending`AnySAMInputFormat` to also override `getSplits` seems to fix the problem, but it's a nasty hack that relies on casting things in various ways.  We should fix it at the source instead.

```
  public static class SplitSortingSamInputFormat extends AnySAMInputFormat{
        @SuppressWarnings("unchecked")
        @Override
        public List<InputSplit> getSplits(JobContext job) throws IOException {
            final List<InputSplit> splits = super.getSplits(job);


            if( splits.stream().allMatch(split -> split instanceof FileVirtualSplit || split instanceof FileSplit)) {
                splits.sort(Comparator.comparing(split -> {
                    if (split instanceof FileVirtualSplit) {
                        return ((FileVirtualSplit) split).getPath();
                    } else {
                        return ((FileSplit) split).getPath();
                    }
                }));
            }

            return splits;
        }
    }
```

We noticed this as part of adding tests on sharded files in https://github.com/broadinstitute/gatk/pull/4545/files.  The tests passed on OSX but failed with out of order files on Travis (running ubuntu).

lbergelson:
@tomwhite Could you give us your thoughts on this when you get a chance?

tomwhite:
I thought that Hadoop's FileInputFormat returned splits in lexicographic order, see

https://hadoop.apache.org/docs/r2.8.2/api/org/apache/hadoop/fs/FileSystem.html#globStatus(org.apache.hadoop.fs.Path)

where it says "Results are sorted by their names". Have you got some more details about the input file path you used, the files actually in the directory, and the order that the splits were read in?

The Hadoop-BAM fix would be to change AnySAMInputFormat to do the sort.

PR #200, opened 2018-06-14
Ensure splits are sorted by path name

tomwhite:
Untested fix for #199. @lbergelson are you able to try this out?

### DIFF
--- a/src/main/java/org/seqdoop/hadoop_bam/AnySAMInputFormat.java
+++ b/src/main/java/org/seqdoop/hadoop_bam/AnySAMInputFormat.java
@@ -24,6 +24,7 @@ package org.seqdoop.hadoop_bam;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


lbergelson:
@tomwhite Would it be possible for you to publish a snapshot of this?   I'm not sure how to publish hadoop bam snapshots and I was only able to reproduce the original problem on travis.

src/main/java/org/seqdoop/hadoop_bam/AnySAMInputFormat.java
```diff
@@ -229,18 +230,21 @@ else if (split instanceof FileVirtualSplit)
 		final List<InputSplit> origSplits =
 				BAMInputFormat.removeIndexFiles(super.getSplits(job));
 
+		final List<InputSplit> sortedSplits = new ArrayList<>(origSplits);
+		sortedSplits.sort(Comparator.comparing(split -> ((FileSplit) split).getPath()));
```
lbergelson: Is this cast problematic?  I ran into the problem that some splits are `FileVirtualSplits` which are not a subtype of `FileSplit`, there isn't any common super type that offers `getPath`.  Maybe `FileVirtualSplit` should extend `FileSplit`?

tomwhite: `AnySAMInputFormat` extends Hadoop's `FileInputFormat`, which returns `FileSplit` objects and never `FileVirtualSplit`. So the cast is safe at this point in the code.

tomwhite:
@lbergelson Snapshots are only published from master, not for PRs, and I don't know a simple way to publish a snapshot.

One way to workaround this is to use a Maven system dependency (just for testing of course). Check in the Hadoop-BAM JAR that you've built locally, then change the dependency in GATK to reference it by path; see an example here: https://github.com/broadinstitute/gatk/compare/tw_squark#diff-c197962302397baf3a4cc36463dce5ea.

magicDGS:
Another way is to use a snapshot built with jitpack.io - https://jitpack.io/#HadoopGenomics/Hadoop-BAM

